Issue #9, opened 2021-10-02
Unify native methods to RAFT


wjxiz1992:
There remains some native functions such as 
https://github.com/NVIDIA/spark-rapids-ml/blob/branch-21.10/native/src/rapidsml_jni.cu#L192
https://github.com/NVIDIA/spark-rapids-ml/blob/branch-21.10/native/src/rapidsml_jni.cu#L271
that are using cusolver functions directly, need to unify all native calls to RAFT APIs.

PR #15, opened 2021-10-11
Unify api using raft

wjxiz1992:
- update cublas calls to raft wrappers
- add docs

fix https://github.com/NVIDIA/spark-rapids-ml/issues/9
Note, the [dgemm_1b](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-21.10/native/src/rapidsml_jni.cu#L224) will be remove after transform(inference) is optimized.

### DIFF
--- a/native/src/rapidsml_jni.cu
+++ b/native/src/rapidsml_jni.cu
@@ -65,16 +65,14 @@ void signFlip(
 
 cublasOperation_t convertToCublasOpEnum(int int_type)
 {
-  if (int_type == 0) {
-    return CUBLAS_OP_N;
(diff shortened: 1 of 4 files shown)


native/src/rapidsml_jni.cu
```diff
@@ -62,6 +62,21 @@ void signFlip(
     }
   });
 }
+
+cublasOperation_t convertToCublasOpEnum(int int_type)
+{
+  if (int_type == 0) {
```
rongou, line 68: Might be better with a `switch` statement.

PR #16, opened 2021-10-12
speedup inference

wjxiz1992:
draft

### DIFF
--- a/src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
+++ b/src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
@@ -187,7 +187,7 @@ class RapidsPCAModel(
           gpuIdBC.value
         }
         val partition = iterator.toList
-        val bas = partition.map(v => v.asBreeze.toArray)
+        val bas = partition.map(v => v.toArray)
         val nvtxRangeConcat = new NvtxRange("concat before transform", NvtxColor.PURPLE)
 
         val A = try {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +171,71 @@ class RapidsPCAModel(
...
+      }
+      val n = input.first().size
+
+      val transformed = input.mapPartitions(iterator => {
+        val gpu = if (gpuIdBC.value == -1) {
+          TaskContext.get().resources()("gpu").addresses(0).toInt
+        } else {
+          gpuIdBC.value
+        }
+        val partition = iterator.toList
```
wjxiz1992, line 189: // maybe data copy

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +171,71 @@ class RapidsPCAModel(
...
+        val bas = partition.map(v => v.asBreeze.toArray)
```
wjxiz1992, line 190: no need asBreeze
```suggestion
        val bas = partition.map(v => v.toArray)
```

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +171,71 @@ class RapidsPCAModel(
...
+        val nvtxRangeConcat = new NvtxRange("concat before transform", NvtxColor.PURPLE)
+
+        val A = try {
+          new DenseMatrix(bas.length, n, Array.concat(bas: _*), isTransposed = true)
```
wjxiz1992, line 194: Consider using unsafe memory copy to replace Array.concat. This step takes about 10 seconds in profiling while GPU computation takes 2 seconds

Issue #17, opened 2021-10-15
Integrate spark-rapids plugin for transform speedup


wjxiz1992:
With some profiling, we found the row concatenation is not efficient. Using spark-rapids plugin to load data to GPU and get ColumnarBatch directly can eliminate the row concatenation time. 

TODO:
1. understand how cuml retrieve native device data from higher level input such as cuDF or cuDF series. (as_gpu_matrix)
2. explore cuDF JNI side for device data access methods.  (ColumnVector.getData.getAddress)
3. prototype and profile ( using RapidsUDF, example: [URLDecode](https://github.com/NVIDIA/spark-rapids/blob/branch-21.12/udf-examples/src/main/scala/com/nvidia/spark/rapids/udf/scala/URLDecode.scala))

steps of implementation:
- create a data buffer(DeviceMemoryBuffer)
- create child column
- create offset column(sequence)
- create the target column

PR #19, opened 2021-10-19
Use spark-rapids plugin to enable columnar process for PCA transform

wjxiz1992:
to fix: https://github.com/NVIDIA/spark-rapids-ml/issues/17


src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,13 +171,34 @@ class RapidsPCAModel(
...
+        require(args.length == 1, s"Unexpected argument count: ${args.length}")
+        val input = args.head
+        val rows_A = input.getRowCount.toInt
+        val AdevAddr = input.getData.getAddress
+        var C: Long = 0
+        // A: raw data, B: pc, C: output, deviceID= 0 for test
+        val lengthInBytes = RAPIDSML.gemm_test(RAPIDSML.CublasOperationT.CUBLAS_OP_N.id,
+          RAPIDSML.CublasOperationT.CUBLAS_OP_N.id,
+          rows_A, pc.numCols, cols_A, 1.0, AdevAddr, cols_A, pc, cols_A, 0.0, C, rows_A, 0)
+        val dmb = new DeviceMemoryBuffer(C, (rows_A*pc.numCols*8).toLong, Cuda.DEFAULT_STREAM)
```
firestarman, line 186: This constructor is accessible inside package `ai.rapids.cudf`.  So you need to put this code under that package.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,13 +171,34 @@ class RapidsPCAModel(
...
+        val childColumn = new ColumnVector(DType.FLOAT64, rows_A.toLong, Optional.of(0), dmb,
+          null, null)
+        val offsetCV = ColumnVector.sequence(Scalar.fromInt(0), Scalar.fromInt(pc.numCols), rows_A)
+        val toClose = List(dmb)
+        val childHandles = Array(childColumn.getNativeView)
+
+        new ColumnVector(DType.LIST, rows_A, Optional.of(0), null, null, offsetCV.getData, toClose, childHandles)
```
firestarman, line 193: `offsetCV.getData` returns a `BaseDeviceMemoryBuffer`, you can use `offsetCV.getData.asInstanceof[DeviceMemoryBuffer]`

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +173,58 @@ class RapidsPCAModel(
...
+        val input = args.head
+        val rows_A = input.getRowCount.toInt
+        val AdevAddr = input.getData.getAddress
+        var C: Long = 0
+        // A: raw data, B: pc, C: output, deviceID= 0 for test
+        C = RAPIDSML.gemm_test(RAPIDSML.CublasOperationT.CUBLAS_OP_N.id,
+          RAPIDSML.CublasOperationT.CUBLAS_OP_N.id,
+          rows_A, pc.numCols, cols_A, 1.0, AdevAddr, cols_A, pc, cols_A, 0.0, C, rows_A, 0)
+        val dmb = buildDeviceMemoryBuffer(C, (rows_A * pc.numCols * DType.FLOAT64.getSizeInBytes).toLong)
+        val childColumn = new ColumnVector(DType.FLOAT64, rows_A.toLong, Optional.of(0), dmb,
```
firestarman, line 193: `childColumn` needs to be closed

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +173,58 @@ class RapidsPCAModel(
...
+          null, null)
+        val offsetCV = ColumnVector.sequence(Scalar.fromInt(0), Scalar.fromInt(pc.numCols), rows_A)
```
firestarman, line 195: `offsetCV` needs to be closed.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +173,58 @@ class RapidsPCAModel(
...
+        val toClose = new java.util.ArrayList[DeviceMemoryBuffer]()
+        toClose.add(dmb)
```
firestarman, line 197: dmb will be taken by childColumn, no need to be closed

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: updated.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: removed from toClose

native/src/rapidsml_jni.cu
```diff
@@ -75,6 +103,17 @@ cublasOperation_t convertToCublasOpEnum(int int_type)
       break;
   }
 }
+
+// TODO: optimize for GPU
+__global__ void transposeData(void* in, void* out, int rows, int cols){
```
jlowe, line 108: I don't see this being referenced?

native/src/rapidsml_jni.cu
```diff
@@ -75,6 +103,17 @@ cublasOperation_t convertToCublasOpEnum(int int_type)
       break;
   }
 }
+
+// TODO: optimize for GPU
```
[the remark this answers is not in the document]
wjxiz1992: Oh, this just reminds me that leveraging `CUBLAS_OP_T` properly and swap the position of A&B can help get rid of the extra transpose. Because when matrix: `A * B = C` , we can get `B(t) * A(t) = C(t)` (t means transpose). I add more comments in the UDF part.

native/src/rapidsml_jni.cu
wjxiz1992: Removed.

native/src/rapidsml_jni.cu
```diff
@@ -178,6 +217,7 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
 
   raft::handle_t raft_handle;
   cudaStream_t stream = raft_handle.get_stream();
+  // raft_handle.get_cublas_handle();
```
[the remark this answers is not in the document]
wjxiz1992: Removed. I will commit a new PR to refactor this gemm call that is used for training when I learn enough things from this transform experience.

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +281,74 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
 }
 
+JNIEXPORT void Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemmWithColumnViewPointer(JNIEnv *env, jclass, jint transa, jint transb, jint m, jint n,
+                                                                                      jint k, jdouble alpha, jlong A, jint lda, jdoubleArray B,
+                                                                                      jint ldb, jdouble beta, jlongArray C, jint ldc, jint deviceID)
+{
+  cudaSetDevice(deviceID);
+  raft::handle_t raft_handle;
+  cudaStream_t stream = raft_handle.get_stream();
+  jclass jlexception = env->FindClass("java/lang/Exception");
```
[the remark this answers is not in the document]
wjxiz1992: Updated.

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +281,74 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+
+  try {
+    cudf::column_view *A_cv_ptr = reinterpret_cast<cudf::column_view *>(A);
+    // init cuda stream view from rmm
+    auto c_stream = rmm::cuda_stream_view(reinterpret_cast<cudaStream_t>(stream));
+
+    auto size_B = env->GetArrayLength(B);
+    auto *host_B = env->GetDoubleArrayElements(B, nullptr);
```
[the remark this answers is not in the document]
wjxiz1992: It seems I should use this RAII wrapper, I didn't find a . Can I just copy this `jni_utils.hpp` files directly to this project so that I can use them directly?

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +281,74 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+    }
+
+    // create offset column
+    auto zero = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
+    zero->set_valid_async(true);
+    using ScalarType = cudf::scalar_type_t<cudf::size_type>;
+    static_cast<ScalarType *>(zero.get())->set_value(0);
+    auto step = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
+    step->set_valid_async(true);
+    static_cast<ScalarType *>(zero.get())->set_value(n);
```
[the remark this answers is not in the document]
wjxiz1992: thank you, updated, it's much simpler.

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +281,74 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+    std::unique_ptr<cudf::column> offset_column = cudf::sequence(m + 1, *zero.get(), *step.get(), rmm::mr::get_current_device_resource());
+
+    auto target_column = cudf::make_lists_column(m, std::move(offset_column), std::move(child_column), 0, rmm::device_buffer());
+
+    auto *host_C = env->GetLongArrayElements(C, nullptr);
+    host_C[0] = reinterpret_cast<jlong>(target_column.release());
```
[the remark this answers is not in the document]
wjxiz1992: Updated.

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +269,64 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+    auto size_A = env->GetArrayLength(A);
+    auto *host_A = env->GetDoubleArrayElements(A, nullptr);
+    rmm::device_buffer dev_buff_A = rmm::device_buffer(host_A, size_A * sizeof(double), c_stream);
+    auto size_C = m * n;
+    //create child column that will own the computation result
+    auto child_column = cudf::make_numeric_column(cudf::data_type{cudf::type_id::FLOAT64}, size_C);
+    auto child_mutable_view = child_column->mutable_view();
+    auto status = raft::linalg::cublasgemm(raft_handle.get_cublas_handle(), convertToCublasOpEnum(transa), convertToCublasOpEnum(transb),
+                                           m, n, k, &alpha, (double *)dev_buff_A.data(), lda, (double *)B_cv_ptr->data<double>(),
+                                           ldb, &beta, (double *)child_mutable_view.data<double>(), ldc, stream);
```
jlowe, line 299: The casts on the column view `data` calls should not be necessary since it's already templated on the type.
```suggestion
                                           m, n, k, &alpha, (double const *)dev_buff_A.data(), lda, B_cv_ptr->data<double>(),
                                           ldb, &beta, child_mutable_view.data<double>(), ldc, stream);
```

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +269,64 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+
+    if (status != CUBLAS_STATUS_SUCCESS) {
+      throw_java_exception(env, RUNTIME_ERROR_CLASS, "cublasgemm failed");
+    }
+
+    // create offset column
+    auto zero = cudf::numeric_scalar<int32_t>(0, true, c_stream);
+	  auto step = cudf::numeric_scalar<int32_t>(m, true, c_stream);
```
jlowe, line 307: Indentation is off here, looks like a case of tabs vs. spaces.
```suggestion
    auto step = cudf::numeric_scalar<int32_t>(m, true, c_stream);
```

src/main/java/com/nvidia/spark/ml/linalg/JniRAPIDSML.java
```diff
@@ -65,7 +65,7 @@ public static JniRAPIDSML getInstance() {
 
   public native void dgemm(int transa, int transb, int m, int n, int k, double alpha, double[] A, int lda, double[] B,
                            int ldb, double beta, double[] C, int ldc, int deviceID);
-  public native void dgemm_b(int m, int n, int k, double[] A, double[] B, double[] C, int deviceID);
-
+  public native long dgemmWithColumnViewPointer(int transa, int transb, int m, int n, int k, double alpha, double[] A,
```
jlowe, line 68: Since this is suppose to be an example, it would probably help to have some Javadoc for this native method explaining what the parameters are.  For example this just says `long B` which may confuse readers who don't realize `B` is the address of a `cudf::column_view *` that represents the B data (transposed).

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +185,61 @@ class RapidsPCAModel(
...
+        val gpu = if (gpuIdBC.value == -1) {
+          TaskContext.get().resources()("gpu").addresses(0).toInt
+        } else {
+          gpuIdBC.value
+        }
+        require(args.length == 1, s"Unexpected argument count: ${args.length}")
+        val input = args.head
+        val rows_A = input.getRowCount.toInt
+        val childCViewNative = input.getChildColumnView(0).getNativeView
+        // Due to the layout of LIST type ColumnVecotr, cublas gemm function should return the transposed result matrix
```
jlowe, line 208:
```suggestion
        // Due to the layout of LIST type ColumnVector, cublas gemm function should return the transposed result matrix
```

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +73,28 @@ private[spark] object RAPIDSML extends Serializable {
     jniRAPIDSML.dgemm(transa, transb, m, n, k, alpha, A.values, lda, B.values, ldb, beta, C.values, ldc, deviceID)
   }
 
+  /**
+   *
```
jlowe, line 77: There's no documentation as to what this does, just a list of parameter descriptions.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +73,28 @@ private[spark] object RAPIDSML extends Serializable {
...
+   * @param transa int type representing the operation op(A) that is non- or (conj.) transpose.
+   * @param transb int type representing the operation op(B) that is non- or (conj.) transpose.
```
jlowe, line 79: This says it's an int type -- what values are expected here?  Curious why this isn't taking CublasOperationT.Value as the parameter type so users know what is accepted and the compiler can help validate.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +73,28 @@ private[spark] object RAPIDSML extends Serializable {
...
+   * @param m number of rows of matrix op(A) and C.
+   * @param n number of columns of matrix op(B) and C.
+   * @param k number of columns of op(A) and rows of op(B).
```
jlowe, line 82: Are these necessary as separate parameters, or can they be derived from other parameters?  For example, `A` is being passed as a `DenseMatrix`, can we simply query that object for its row count to get the `m` parameter?

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +73,28 @@ private[spark] object RAPIDSML extends Serializable {
...
+   * @param alpha scalar used for multiplication.
+   * @param A array of dimensions lda x k with lda>=max(1,m) if transa == CUBLAS_OP_N and lda x m with lda>=max(1,k) otherwise.
+   * @param lda leading dimension of two-dimensional array used to store the matrix A.
+   * @param B array of dimension ldb x n with ldb>=max(1,k) if transb == CUBLAS_OP_N and ldb x k with ldb>=max(1,n) otherwise.
```
jlowe, line 86: This says it's an array but it's actually a long.  This will need to elaborate more to help user follow the example.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +73,28 @@ private[spark] object RAPIDSML extends Serializable {
...
 
+  /**
+   *
+   * @param transa int type representing the operation op(A) that is non- or (conj.) transpose.
+   * @param transb int type representing the operation op(B) that is non- or (conj.) transpose.
+   * @param m number of rows of matrix op(A) and C.
+   * @param n number of columns of matrix op(B) and C.
+   * @param k number of columns of op(A) and rows of op(B).
+   * @param alpha scalar used for multiplication.
+   * @param A array of dimensions lda x k with lda>=max(1,m) if transa == CUBLAS_OP_N and lda x m with lda>=max(1,k) otherwise.
```
jlowe, line 84: Says it's an array but it's a `DenseMatrix`.

jlowe (commented):
> Can I just copy this jni_utils.hpp files directly to this project so that I can use them directly?

Isn't this project pulling in cudf and building libcudf?  The header should already be available in the `java/src/main/native/include/` directory in the cudf repo pulled in by the libcudf build.  IMO that would be better than copying, as the copy will get out of date as we make fixes/enhancements to it in cudf.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: Oops, corrected.

src/main/java/com/nvidia/spark/ml/linalg/JniRAPIDSML.java
wjxiz1992: Updated.

native/src/rapidsml_jni.cu
wjxiz1992: Updated.

native/src/rapidsml_jni.cu
wjxiz1992: Added.

native/src/rapidsml_jni.cu
```diff
@@ -247,16 +269,64 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
+JNIEXPORT jlong Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemmWithColumnViewPointer(JNIEnv *env, jclass, jint transa, jint transb, jint m, jint n,
+                                                                                      jint k, jdouble alpha, jdoubleArray A, jint lda, jlong B,
+                                                                                      jint ldb, jdouble beta, jint ldc, jint deviceID)
+{
+  cudaSetDevice(deviceID);
+  raft::handle_t raft_handle;
+  cudaStream_t stream = raft_handle.get_stream();
+
+  try {
+    cudf::column_view *B_cv_ptr = reinterpret_cast<cudf::column_view *>(B);
```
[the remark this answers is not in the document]
wjxiz1992: Added

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
wjxiz1992: Added.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
wjxiz1992: Updated to use CublasOperationT.Value. It's more intuitive.
The original motivation is to avoid passing enum object to JNI directly, and use int as a bridge value. As the GEMM routine need an enum value as well but in native. So I added a [conversion function](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-21.12/native/src/rapidsml_jni.cu#L66-L77 ) to do the job.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
wjxiz1992: Yes, and the most original GEMM wrapper were only using DenseMatrix as input and query other parameters from it. 
But then we were thinking to align with RAFT APIs, so we proposed this PR: https://github.com/NVIDIA/spark-rapids-ml/pull/15, to make it more “familiar” for users who know GEMM well. 
(Actually from my feeling, GEMM parameters are really really confusing, at least for me at the beginning. during my implementation, I spent 20% time finding proper methods to construct column and 80% time thinking how the matrix should multiply with each other and filling proper parameters into this GEMM function. )


So, for GEMM users, I think "familar" parameters are better. For a freshman like me who has no gemm experience, I prefer the "less" parameters way. What do you think?

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
wjxiz1992: Updated.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
jlowe: We could supply both versions, one that is simplified with minimal parameters and one that looks like the full GEMM interface.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +74,34 @@ private[spark] object RAPIDSML extends Serializable {
...
+   * @param transa CublasOperationT enum value representing the operation op(A) that is non- or (conj.) transpose.
+   * @param transb CublasOperationT enum value representing the operation op(B) that is non- or (conj.) transpose.
+   * @param m number of rows of matrix op(A) and C.
+   * @param n number of columns of matrix op(B) and C.
+   * @param k number of columns of op(A) and rows of op(B).
+   * @param alpha scalar used for multiplication.
+   * @param A array of dimensions lda x k with lda>=max(1,m) if transa == CUBLAS_OP_N and lda x m with lda>=max(1,k) otherwise.
+   * @param lda leading dimension of two-dimensional array used to store the matrix A.
+   * @param B long value representing the handle of a `cudf::column_view *` which is holding the matrix data in device.
+   *          (Array of dimension ldb x n with ldb>=max(1,k) if transb == CUBLAS_OP_N and ldb x k with ldb>=max(1,n) otherwise.)
```
jlowe, line 92: Rather than take such a "raw" interface as a `cudf::column_view *`, this should take a `ColumnView` for `B`.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -16,8 +16,8 @@
 
 package org.apache.spark.ml.feature
```
jlowe, line 17: Not part of this PR, but ideally this should not be directly in the Apache Spark package.  We should try as much as possible for this to be in a `com.nvidia.spark....` package.  If something needs to stay under the Apache Spark package then we should minimize its footprint as much as possible and put it under a new package within the Apache Spark package (e.g.: org.apache.spark.ml.feature.rapids)

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +185,61 @@ class RapidsPCAModel(
...
+        logDebug("==========using GPU transform==========")
+        val gpu = if (gpuIdBC.value == -1) {
+          TaskContext.get().resources()("gpu").addresses(0).toInt
+        } else {
+          gpuIdBC.value
+        }
+        require(args.length == 1, s"Unexpected argument count: ${args.length}")
+        val input = args.head
+        val rows_A = input.getRowCount.toInt
+        val childCViewNative = input.getChildColumnView(0).getNativeView
```
jlowe, line 207: `getChildColumnView` returns a native resource in host memory and is leaked here since it is never closed.  This needs to track the `ColumnView` returned in a try..finally block to ensure it is always closed.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
jlowe: Was it corrected?  It still looks the same.  Was a recent commit not pushed?

src/main/java/com/nvidia/spark/ml/linalg/JniRAPIDSML.java
jlowe: Still looks the same,  was a commit not pushed?

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: Oh, sorry, it seems IDEA doesn't save the change. I'll confirm and submit along with the rest issues later. Thank you for your patience.

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
wjxiz1992: Updated, it's now taking ColumnView as a parameter.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: Noted.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
wjxiz1992: Updated, now pass `input`(parent ColumnView) to native directly.

native/src/CMakeLists.txt
```diff
@@ -38,14 +41,40 @@ else(DEFINED ENV{RAFT_PATH})
...
+
+include(${CPM_DOWNLOAD_LOCATION})
+#################################################################################################
+
+# pull cuDF sources, to use jni_utils.hpp
+# cmake options should be added here for CI build.
+CPMAddPackage(NAME cudf
+        VERSION         "21.12.00"
+        GIT_REPOSITORY  https://github.com/rapidsai/cudf.git
+        GIT_TAG         branch-21.12
```
wjxiz1992, line 65: no cmake options because I have installed cuDF via Conda, it's used for local dev only. Will add options here for official CI.

native/src/rapidsml_jni.cu
```diff
@@ -28,12 +28,41 @@
...
 #include <nvtx3.hpp>
+#include <stdlib.h>
+#include <cudf/column/column.hpp>
+#include <cudf/column/column_view.hpp>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/sequence.hpp>
+#include <cudf/lists/lists_column_view.hpp>
+#include <cudf/scalar/scalar_factories.hpp>
+#include <cudf/utilities/type_dispatcher.hpp>
+// #include "jni_utils.hpp"
```
wjxiz1992, line 39: compile error when make
```
/home/allxu/code/spark-rapids-ml/target/native/_deps/cudf-src/java/src/main/native/include/jni_utils.hpp(331): error: reinterpret_cast cannot cast away const or other type qualifiers
```

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +185,56 @@ class RapidsPCAModel(
...
+        // its memory data layout from Cublas GEMM is [1,3,5,2,4,6]. However, it will be displayed in LIST ColumnVector as :
+        // [1,3]
+        // [5,2]
+        // [4,6]
+        // To fill the gap between native memory and CV data storage, we consider the following nature:
+        // when A * B = C, we get BT * AT = CT. (T means transpose). In this case the output matrix becomes:
+        // [1,3,5]
+        // [2,4,6]
+        // whose memory data layout is [1,2,3,4,5,6]. Then it can be comsumed by CV directly.
+        val C = RAPIDSML.gemmWithColumnViewPointer(pc, input, input_rows ,gpu)
```
jlowe, line 219: I like how much simpler this call now looks!  Is it necessary to pass `input_rows` here since it can be trivially derived from `input`?  Also the name is a bit out of date with the recent refactoring, should be updated to something like `gemmWithColumnView` or maybe just `gemm`.

src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -168,26 +185,56 @@ class RapidsPCAModel(
...
+      }
+
+      override def apply(v1: mutable.WrappedArray[Double]): Array[Double] = {
+        logDebug("==========using CPU transform==========")
+        pc.transpose.multiply(Vectors.dense(v1.toArray)).toArray
+      }
+    }
+
+    if (getUseGemm) {
+      val transform_udf = dataset.sparkSession.udf.register("transform", new gpuTransform())
```
jlowe, line 230: Do we risk colliding with another instance of this class or even a user's UDF?  `transform` is a pretty generic name that some user could have already used for another UDF.

src/main/java/com/nvidia/spark/ml/linalg/JniRAPIDSML.java
```diff
@@ -65,7 +65,7 @@ public static JniRAPIDSML getInstance() {
 
   public native void dgemm(int transa, int transb, int m, int n, int k, double alpha, double[] A, int lda, double[] B,
                            int ldb, double beta, double[] C, int ldc, int deviceID);
-  public native void dgemm_b(int m, int n, int k, double[] A, double[] B, double[] C, int deviceID);
-
+  public native long dgemm(int transa, int transb, int m, int n, int k, double alpha, double[] A, int lda, long B,
```
jlowe, line 68: Sorry, my mistake.  When I suggested for this to be renamed `dgemm` I didn't realize it was a native function where overloading is problematic.  Now I think it's worse than before, because:
- There's no Javadoc for any of this.  How does one know what the parameters mean and why `B` can be a `long` when it represents a matrix?
- Overloaded methods in native code are not so nice, as you discovered with the mangled name required in native code.

native/src/ml_utils.cu
```diff
@@ -0,0 +1,85 @@
+#include "ml_utils.hpp"
```
jlowe, line 1: Need NVIDA copyrights on this and the other newly added files.

native/src/rapidsmlJniRemain.cu
```diff
@@ -0,0 +1,149 @@
+#include <cuda_runtime.h>
```
jlowe, line 1: The `Remain` name in the file seems hacky and non-descript, and I'm a bit surprised to see the rapidsml_jni.cu file deleted.  Why not leave rapidsml_jni.cu in place and add rapmdsmlJni.cpp with the new dgemm stuff?  That would be a much smaller diff for this PR.

native/src/ml_utils.cu
wjxiz1992: Added, also create an issue to add rat plugin to avoid this. https://github.com/NVIDIA/spark-rapids-ml/issues/22

native/src/rapidsmlJniRemain.cu
wjxiz1992: Agree, renamed them back.

native/src/rapidsmlJni.cpp
```diff
@@ -0,0 +1,56 @@
...
+               (e.what() == nullptr ? "" : e.what());
+    cudf::jni::throw_java_exception(env, RUNTIME_ERROR_CLASS, msg.c_str());
+  }
+  catch (std::invalid_argument const &e) {
+    cudf::jni::throw_java_exception(env, cudf::jni::ILLEGAL_ARG_CLASS, e.what() == nullptr ? "" : e.what());
+  }
+  catch (std::exception const &e) {
+    auto msg = e.what() == nullptr ? "" : e.what();
+    cudf::jni::throw_java_exception(env, RUNTIME_ERROR_CLASS, msg);
+  }
```
jlowe, line 54: Can this just use the `CATCH_STD` macro from jni_utils.hpp here?

native/src/rapidsml_jni.cu
```diff
@@ -75,99 +75,39 @@ cublasOperation_t convertToCublasOpEnum(int int_type)
...
-JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_NvtxRange_pop(JNIEnv *env, jclass clazz) {
-  jclass jlexception = env->FindClass("java/lang/Exception");
-  try {
-    nvtxDomainRangePop(nvtx3::domain::get<java_domain>());
-  } catch (const std::bad_alloc &e) {
-    env->ThrowNew(jlexception, "Error nvtx pop");
-  }
-  // CATCH_STD(env, );
+long dgemm(int transa, int transb, int m, int n,
+                            int k, double alpha, double* A, int size_A, int lda, long B,
```
jlowe, line 80: Nit: Indentation is odd here

native/src/rapidsml_jni.cu
```diff
@@ -213,13 +153,7 @@ JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniRAPIDSML_dgemm(JNIEnv*
...
 
-  cublasHandle_t handle;
-  auto status = cublasCreate(&handle);
-  if (status != CUBLAS_STATUS_SUCCESS) {
-    env->ThrowNew(jlexception, "Error creating cuBLAS handle");
-  }
-
-  status = raft::linalg::cublasgemm(handle, convertToCublasOpEnum(transa), convertToCublasOpEnum(transb), m, n, k, &alpha, dev_A, lda, dev_B, ldb, &beta,
+  auto status = raft::linalg::cublasgemm(raft_handle.get_cublas_handle(), convertToCublasOpEnum(transa), convertToCublasOpEnum(transb), m, n, k, &alpha, dev_A, lda, dev_B, ldb, &beta,
                        dev_C, ldc, stream);
```
jlowe, line 157: Nit: Indentation is odd here

native/src/rapidsml_jni.hpp
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <raft/linalg/cublas_wrappers.h>
+
+namespace {
```
jlowe, line 21: [Anonymous namespaces in C++ headers](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file#:~:text=an%20%23include%20directive.-,Do%20not%20define%20an%20unnamed%20namespace%20in%20a%20header%20file,used%20within%20that%20translation%20unit.) is an anti-pattern.

src/main/java/com/nvidia/spark/ml/linalg/JniRAPIDSML.java
```diff
@@ -61,11 +61,16 @@ public static JniRAPIDSML getInstance() {
...
-  public native void dgemm_b(int m, int n, int k, double[] A, double[] B, double[] C, int deviceID);
 
+  /** Wrapper of JNI entrance for cuBLAS gemm routine. Most parameters are the same as the original gemm's: https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemm.
+   * Differences are:
+   * 1. transa and transb are int values instead of enum.
+   * 2. B is a long value that represeents the `cudf::lists_column_view *` holding the matrix data on device
+   * 3. an extra deviceID to indicate which GPU device will perform this computation
+   */
+  public native long dgemmWithColumnViewPtr(int transa, int transb, int m, int n, int k, double alpha, double[] A, int lda, long B,
+                           int ldb, double beta, int ldc, int deviceID);
```
jlowe, line 74: Nit: Indentation is odd here

src/main/scala/org/apache/spark/ml/linalg/RAPIDSML.scala
```diff
@@ -73,6 +66,52 @@ private[spark] object RAPIDSML extends Serializable {
...
+   * @param ldb leading dimension of two-dimensional array used to store matrix B.
+   * @param beta scalar used for multiplication. If beta==0, C does not have to be a valid input.
+   * @param ldc leading dimension of a two-dimensional array used to store the matrix C.
+   * @param deviceID the device that will run the computation
+   * @return value of Long type that represents the handle of a LIST type `cudf::column_view *` which is holding the computation output.
+   *         It can be used to construct ColumnVector.
+   */
+  def gemm(transa: CublasOperationT, transb: CublasOperationT, m: Int, n: Int, k: Int, alpha: Double, A: Array[Double],
+                                lda: Int, B: ColumnView, ldb: Int,beta: Double, ldc: Int, deviceID: Int): Long = {
```
jlowe, line 95: Nit: Indentation is odd here

jlowe (commented):
This looks pretty close.  Mostly minor nits remain.

native/src/rapidsmlJni.cpp
wjxiz1992: Updated.

native/src/rapidsml_jni.hpp
wjxiz1992: Removed, and url added to bookmark 👍

jlowe (approved):
Assuming this builds and passes tests, lgtm.

native/src/rapidsmlJni.cpp
```diff
@@ -0,0 +1,45 @@
+/*
```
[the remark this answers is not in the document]
wjxiz1992: Oh, yes, Let me change it back. I was just trying to mimic the naming strategy in udf-example at that time: https://github.com/NVIDIA/spark-rapids/tree/branch-21.12/udf-examples/src/main/cpp/src .

rongou (approved):

Issue #24, opened 2021-11-12
[BUG] build failure due to RAFT header change


wjxiz1992:
In this PR: https://github.com/rapidsai/raft/pull/356, raft has changed lots of header file names. We need to modify accordingly.

```
20:18:11  [INFO]      [exec] /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -DCUDA_API_PER_THREAD_DEFAULT_STREAM -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -Drapidsml_jni_EXPORTS -I/rapids/rmm/include/rmm -I/usr/lib/jvm/java-1.8.0-openjdk/include -I/usr/lib/jvm/java-1.8.0-openjdk/include/linux -I/rapids/raft/cpp/include -I/home/jenkins/agent/workspace/jenkins-rapids-ml-nightly-57-cuda115/native/src/utils -isystem=/usr/local/cuda/include --extended-lambda --generate-code=arch=compute_52,code=[compute_52,sm_52] -Xcompiler=-fPIC -std=c++17 -MD -MT src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cu.o -MF src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cu.o.d -x cu -c /home/jenkins/agent/workspace/jenkins-rapids-ml-nightly-57-cuda115/native/src/rapidsml_jni.cu -o src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cu.o
20:18:11  [INFO]      [exec] /home/jenkins/agent/workspace/jenkins-rapids-ml-nightly-57-cuda115/native/src/rapidsml_jni.cu:25:10: fatal error: raft/matrix/matrix.cuh: No such file or directory
20:18:11  [INFO]      [exec]    25 | #include <raft/matrix/matrix.cuh>
20:18:11  [INFO]      [exec]       |          ^~~~~~~~~~~~~~~~~~~~~~~~
20:18:11  [INFO]      [exec] compilation terminated.
```

cjnolet:
I noticed this issue references recent API changes in RAFT so I wanted to provide a brief note on the current plan to help us stay in alignment. RAFT is currently in the process of exposing a formal public API and hiding irrelevant implementation details from the new APIs.

[...]

PR #25, opened 2021-11-12
Change raft header file names

wjxiz1992:
- Update header file names to latest raft change
local test passes.
fix https://github.com/NVIDIA/spark-rapids-ml/issues/24


pxLi (approved):
LGTM, not sure if we would like to target dep repo to a specific commit/tag to avoid issue like this

wjxiz1992:
> LGTM, not sure if we would like to target dep repo to a specific commit/tag to avoid issue like this

That's a good point! I checked raft's tags, there's no tag that includes their latest change. I think we can specify a tag after 21.12 release.

PR #36, opened 2021-12-10
Build cudf library locally during the building process

wjxiz1992:
- update CMakeLists.txt to locally build cuDF to get rid of the dependency of arrow library. Previously we installed cuDF via conda but the libcudf there has dynamic link to libarrow which may cause lib not found problem in another runtime.

### DIFF
--- a/native/src/CMakeLists.txt
+++ b/native/src/CMakeLists.txt
@@ -47,10 +47,10 @@ endif(DEFINED ENV{RAFT_PATH})
 set(CUDA_USE_STATIC_CUDA_RUNTIME OFF)
 
 rapids_cpm_init()
-rapids_cpm_find(cudf 22.02.00
+rapids_cpm_find(cudf 21.12.00
         CPM_ARGS
         GIT_REPOSITORY  https://github.com/rapidsai/cudf.git
-        GIT_TAG         branch-22.02
+        GIT_TAG         branch-21.12
         GIT_SHALLOW     TRUE
         SOURCE_SUBDIR   cpp
         OPTIONS         "BUILD_TESTS OFF"
--- a/native/src/CMakeLists.txt
+++ b/native/src/CMakeLists.txt
@@ -16,7 +16,7 @@
 
 
 # Install cuDF nightly via Conda, only for local development, will remove in CI.
-find_package(cudf)
+# find_package(cudf)
 
 set (CMAKE_CUDA_FLAGS "--extended-lambda")
 
@@ -40,29 +40,29 @@ else(DEFINED ENV{RAFT_PATH})
   set(RAFT_INCLUDE_DIR ${RAFT_GIT_DIR}/src/raft/cpp/include CACHE STRING "RAFT include variable")
 endif(DEFINED ENV{RAFT_PATH})
 
+############################################################################################
+# - cudf -----------------------------------------------------------------------------------
 
+# Ensure CUDA runtime is dynamic despite statically linking Arrow in libcudf
+set(CUDA_USE_STATIC_CUDA_RUNTIME OFF)
 
-#################################################################################################
-# - CPM -----------------------------------------------------------------------------------------
-
-set(CPM_DOWNLOAD_VERSION 0.27.2)
-set(CPM_DOWNLOAD_LOCATION "${CMAKE_BINARY_DIR}/cmake/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
-
-if(NOT (EXISTS ${CPM_DOWNLOAD_LOCATION}))
-    message(STATUS "Downloading CPM.cmake")
-    file(DOWNLOAD https://github.com/TheLartians/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake ${CPM_DOWNLOAD_LOCATION})
-endif()
-
-include(${CPM_DOWNLOAD_LOCATION})
-#################################################################################################
-
-# pull cuDF sources, to use jni_utils.hpp
-# cmake options should be added here for CI build.
-CPMAddPackage(NAME cudf
-        VERSION         "21.12.00"
+rapids_cpm_init()
+rapids_cpm_find(cudf 22.02.00
+        CPM_ARGS
         GIT_REPOSITORY  https://github.com/rapidsai/cudf.git
-        GIT_TAG         branch-21.12
-)
+        GIT_TAG         branch-22.02
+        GIT_SHALLOW     TRUE
+        SOURCE_SUBDIR   cpp
+        OPTIONS         "BUILD_TESTS OFF"
+                        "BUILD_BENCHMARKS OFF"
+                        "CUDF_USE_ARROW_STATIC ON"
+                        "JITIFY_USE_CACHE ON"
+                        "CUDA_STATIC_RUNTIME OFF"
+                        "DISABLE_DEPRECATION_WARNING ON"
+                        "AUTO_DETECT_CUDA_ARCHITECTURES OFF"
+    )
+
+############################################################################################
 
 add_library(rapidsml_jni SHARED rapidsml_jni.cpp
                                 rapidsml_jni.cu
--- a/native/src/CMakeLists.txt
+++ b/native/src/CMakeLists.txt
@@ -14,10 +14,6 @@
 # limitations under the License.
 #=============================================================================
 
-
-# Install cuDF nightly via Conda, only for local development, will remove in CI.
-# find_package(cudf)
-
 set (CMAKE_CUDA_FLAGS "--extended-lambda")
 
 if(DEFINED ENV{RAFT_PATH})
--- a/native/src/CMakeLists.txt
+++ b/native/src/CMakeLists.txt
@@ -56,6 +56,7 @@ rapids_cpm_find(cudf 21.12.00
         OPTIONS         "BUILD_TESTS OFF"
                         "BUILD_BENCHMARKS OFF"
                         "CUDF_USE_ARROW_STATIC ON"
+                        "CUDF_ENABLE_ARROW_S3 OFF"
                         "JITIFY_USE_CACHE ON"
                         "CUDA_STATIC_RUNTIME OFF"
                         "DISABLE_DEPRECATION_WARNING ON"
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,6 @@
         <test.include.tags></test.include.tags>
         <cuda.version>cuda11</cuda.version>
         <GPU_ARCHS>ALL</GPU_ARCHS>
-        <CUDF_ENABLE_ARROW_S3>OFF</CUDF_ENABLE_ARROW_S3>
         <CPP_PARALLEL_LEVEL>4</CPP_PARALLEL_LEVEL>
     </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,9 @@
         <test.exclude.tags></test.exclude.tags>
         <test.include.tags></test.include.tags>
         <cuda.version>cuda11</cuda.version>
-
+        <GPU_ARCHS>ALL</GPU_ARCHS>
+        <CUDF_ENABLE_ARROW_S3>OFF</CUDF_ENABLE_ARROW_S3>
+        <CPP_PARALLEL_LEVEL>4</CPP_PARALLEL_LEVEL>
     </properties>
 
     <repositories>
@@ -362,10 +364,15 @@
                         <configuration>
                             <target>
                                 <exec executable="cmake" failonerror="true">
-                                    <arg line="-S ${basedir}/native/ -B ${project.build.directory}/native -GNinja"/>
+                                    <arg value="-S${basedir}/native/"/>
+                                    <arg value="-B${project.build.directory}/native"/>
+                                    <arg value="-DGPU_ARCHS=${GPU_ARCHS}"/>
+                                    <arg value="-DCUDF_ENABLE_ARROW_S3=${CUDF_ENABLE_ARROW_S3}"/>
+                                    <arg value="-GNinja"/>
                                 </exec>
                                 <exec executable="ninja" dir="${project.build.directory}/native" failonerror="true">
-                                    <arg line="-v -j 2"/>
+                                    <arg value="-v"/>
+                                    <arg value="-j${CPP_PARALLEL_LEVEL}"/>
                                 </exec>
                             </target>
                         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -366,7 +366,6 @@
                                     <arg value="-S${basedir}/native/"/>
                                     <arg value="-B${project.build.directory}/native"/>
                                     <arg value="-DGPU_ARCHS=${GPU_ARCHS}"/>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -77,7 +77,9 @@
         <test.exclude.tags></test.exclude.tags>
         <test.include.tags></test.include.tags>
         <cuda.version>cuda11</cuda.version>
-
+        <GPU_ARCHS>ALL</GPU_ARCHS>
+        <CUDF_ENABLE_ARROW_S3>OFF</CUDF_ENABLE_ARROW_S3>
```
jlowe, line 81: Do we really want this to ever be `ON`?  This is just to build libcudf so we can link against it, and the libcudf API signatures don't change based on this setting.  IMO this should be hardcoded in the rapids_cpm_find as we do for ARROW_STATIC, etc.

native/src/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 
 
 # Install cuDF nightly via Conda, only for local development, will remove in CI.
-find_package(cudf)
+# find_package(cudf)
```
jlowe, line 19: Why not simply remove this?

jlowe:
I'm a bit confused on the reported problem.  Is libcudf.so being bundled with a spark-rapids-ml jar or is it supposed to be picking up the libcudf.so that is loaded by the cudf jar?  If the latter, then I don't understand why it matters whether libcudf is linked dynamically or statically against the Arrow library if this module only links against libcudf.so at build time.  Is the spark-rapids-ml code leveraging Arrow APIs and expecting to pick them up through the libcudf.so dependency?

wjxiz1992:
Before this change, we set up the build environment by installing conda, then installing cuDF via conda, and use the libcudf.so from conda, not link the libcudf.so from the cuDF jar.  In this case, after building the spark-rapids-ml jar, we can see some dynamic links like

```
+ ldd ./amd64/Linux/librapidsml_jni.so
ldd: warning: you do not have execution permission for `./amd64/Linux/librapidsml_jni.so'
        linux-vdso.so.1 (0x00007ffc7dba7000)
        libcudf.so => /root/miniconda3/lib/libcudf.so (0x00007ff6cf328000)
        libcublas.so.11 => /usr/local/cuda/lib64/libcublas.so.11 (0x00007ff6c5a90000)
        libarrow.so.500 => /root/miniconda3/lib/libarrow.so.500 (0x00007ff6c48a8000)
        libarrow_cuda.so.500 => /root/miniconda3/lib/libarrow_cuda.so.500 (0x00007ff6f09db000)
```

So if the runtime environemnt doesn't have the libarrow.so, it will crash. There's no use of Arrow APIs but somehow it's linking it.

So I think the only target is to get rid of the libarrow.so

native/src/CMakeLists.txt
wjxiz1992: The only benefit is, when debugging with my conda environment where libcudf is installed, I can use this and comment the "cuDF building part" in this file, so I don't need to wait for about 1 or 2 hours to wait.

pom.xml
wjxiz1992: Ah, yes, that's better. Updated.

native/src/CMakeLists.txt
wjxiz1992: But yes, should make it clean for public read. Updated.

jlowe:
> If the runtime environment doesn't have the libarrow.so, it will crash. There's no use of Arrow APIs but somehow it's linking it.

Note that `ldd` is not necessarily going to tell the true story of what will happen at runtime.  `ldd` will show you all of the dependencies, even if they're transitive.  For example, here's libcudfjni.so and libcudf.so from the latest cudf SNAPSHOT jar:
```
$ ldd amd64/Linux/*
amd64/Linux/libcudfjni.so:
	linux-vdso.so.1 (0x00007ffd48994000)
	libcudf.so => not found
amd64/Linux/libcudf.so:
	linux-vdso.so.1 (0x00007ffc18d03000)
	libcuda.so.1 => /lib/x86_64-linux-gnu/libcuda.so.1 (0x00007f716f021000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f716f016000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f716eff3000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f716efed000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f716ee0b000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f716ecba000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f716ec9f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f716eaad000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f718d6f7000)
```
Note how libcudfjni.so has only a dependency on libcudf.so which is not found.  But if I allow ldd to locate it, notice how it suddenly reports that it needs a _lot_ more libraries:
```
$ LD_LIBRARY_PATH=amd64/Linux ldd amd64/Linux/*
amd64/Linux/libcudfjni.so:
	linux-vdso.so.1 (0x00007ffc19d2c000)
	libcudf.so => amd64/Linux/libcudf.so (0x00007f3228bae000)
	libcuda.so.1 => /lib/x86_64-linux-gnu/libcuda.so.1 (0x00007f32274b1000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f32274a6000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3227483000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f322747d000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f3227299000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f322714a000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f322712f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3226f3d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3245b8c000)
amd64/Linux/libcudf.so:
	linux-vdso.so.1 (0x00007ffe5a7f3000)
	libcuda.so.1 => /lib/x86_64-linux-gnu/libcuda.so.1 (0x00007fea8faa2000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fea8fa97000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fea8fa74000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fea8fa6e000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fea8f88c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fea8f73b000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fea8f720000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fea8f52e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007feaae178000)
```

So it would be interesting to "hide" libcudf.so from librapidsml_jni.so (e.g.: temporarily move it to libcudf.so.moved or something) and see if `ldd` still reports the arrow dependency on spark-rapids-ml.  If it does not show the arrow dependency in that case, then I don't think there's anything to worry about here.  The libcudf.so it finds at runtime (which comes from the cudf jar) will always have a statically linked Arrow library (well, at least if it's built with the Java CI script which builds the "run anywhere" version of the libs).

If `ldd` still shows the Arrow dependency even when libcudf.so cannot be found then we should add static arrow linking to the spark-rapids-ml build, not necessarily build all of libcudf with static arrow and hope that it pulled in the pieces of Arrow that we also need in spark-rapids-ml.

Bottom line is that it seems like this PR isn't the right fix for the problem being reported.  Either there isn't a real problem here at runtime (it just appears that way from `ldd` output) or we need to statically link Arrow into spark-rapids-ml.

rongou (approved):

pom.xml
```diff
@@ -362,10 +363,15 @@
                         <configuration>
                             <target>
                                 <exec executable="cmake" failonerror="true">
-                                    <arg line="-S ${basedir}/native/ -B ${project.build.directory}/native -GNinja"/>
+                                    <arg value="-S${basedir}/native/"/>
+                                    <arg value="-B${project.build.directory}/native"/>
+                                    <arg value="-DGPU_ARCHS=${GPU_ARCHS}"/>
+                                    <arg value="-DCUDF_ENABLE_ARROW_S3=${CUDF_ENABLE_ARROW_S3}"/>
```
jlowe, line 369: This property was removed.
```suggestion
```

wjxiz1992:
@jlowe There's a very confusing observation, when I build the lib in ubuntu, then I rename the libcudf.so in miniconda, I can see:
```
allxu@allxu-System-Product-Name:~/code/spark-rapids-ml/target/native/lib$ ldd librapidsml_jni.so

        linux-vdso.so.1 (0x00007ffd6c357000)
        libcudf.so => not found
        libcublas.so.11 => /usr/local/cuda/lib64/libcublas.so.11 (0x00007f9077eb4000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9077eae000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9077e8b000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9077e80000)
        libstdc++.so.6 => /home/allxu/miniconda3/lib/libstdc++.so.6 (0x00007f9077d0b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9077bba000)
        libgcc_s.so.1 => /home/allxu/miniconda3/lib/libgcc_s.so.1 (0x00007f9077ba6000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f90779b4000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f90848b9000)
        libcublasLt.so.11 => /usr/local/cuda/lib64/libcublasLt.so.11 (0x00007f9063449000)

```

However, when I started a Docker for Centos7 and build in the exact same fasion(mvn package), I got
```
(base) [root@9733387dc6e5 lib]# ldd librapidsml_jni.so
        linux-vdso.so.1 =>  (0x00007ffef61b3000)
        libcudf.so => not found
        libcublas.so.11 => /usr/local/cuda/lib64/libcublas.so.11 (0x00007f35c07fd000)
        libarrow.so.500 => not found
        libarrow_cuda.so.500 => not found
        libdl.so.2 => /usr/lib64/libdl.so.2 (0x00007f35c05f9000)
        libpthread.so.0 => /usr/lib64/libpthread.so.0 (0x00007f35c03dd000)
        libcudart.so.11.0 => /usr/local/cuda/lib64/libcudart.so.11.0 (0x00007f35c013b000)
        libcuda.so.1 => /usr/lib64/libcuda.so.1 (0x00007f35be971000)
        librt.so.1 => /usr/lib64/librt.so.1 (0x00007f35be769000)
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00007f35be461000)
        libm.so.6 => /usr/lib64/libm.so.6 (0x00007f35be15f000)
        libgcc_s.so.1 => /usr/lib64/libgcc_s.so.1 (0x00007f35bdf49000)
        libc.so.6 => /usr/lib64/libc.so.6 (0x00007f35bdb7b000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f35cd1f0000)
        libcublasLt.so.11 => /usr/local/cuda/lib64/libcublasLt.so.11 (0x00007f35a9610000)

```

here's the compile command generated: 
Ubuntu: 
```
&& /usr/bin/c++ -fPIC   
 -shared -Wl,-soname,librapidsml_jni.so -o lib/librapidsml_jni.so src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cpp.o src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cu.o
 -L/usr/local/cuda/targets/x86_64-linux/lib/stubs   -L/usr/local/cuda/targets/x86_64-linux/lib -Wl,
-rpath,/usr/local/cuda/lib64:/home/allxu/miniconda3/lib
  -Wl,-Bstatic  -lcudart_static  -lcusparse_static  -lcusolver_static  -lculibos  -llapack_static 
 -Wl,-Bdynamic  /home/allxu/miniconda3/lib/libcudf.so  /usr/local/cuda/lib64/libcublas.so  /home/allxu/miniconda3/lib/libarrow.so.500.0.0  /home/allxu/miniconda3/lib/libarrow_cuda.so.500.0.0  
-ldl  -lpthread  /usr/local/cuda/lib64/libcudart.so  /usr/lib/x86_64-linux-gnu/libcuda.so 
 -lcudadevrt  -lcudart_static  -lrt  -lpthread  -ldl && 
```

Centos:
```
&& /opt/rh/devtoolset-9/root/usr/bin/c++ -fPIC    
-shared -Wl,-soname,librapidsml_jni.so -o lib/librapidsml_jni.so src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cpp.o src/CMakeFiles/rapidsml_jni.dir/rapidsml_jni.cu.o 
-L/usr/local/cuda/targets/x86_64-linux/lib -Wl,
-rpath,/usr/local/cuda/lib64:/root/miniconda3/lib  
-Wl,-Bstatic  -lcudart_static  -lcusparse_static  -lcusolver_static  -lculibos  -llapack_static  
-Wl,-Bdynamic  /root/miniconda3/lib/libcudf.so  /usr/local/cuda/lib64/libcublas.so  /root/miniconda3/lib/libarrow.so.500.0.0  /root/miniconda3/lib/libarrow_cuda.so.500.0.0  
-ldl  -lpthread  /usr/local/cuda/lib64/libcudart.so  /usr/lib64/libcuda.so 
 -lcudadevrt  -lcudart_static  -lrt  -lpthread  -ldl && 
```

jlowe:
Note that in both cases, the compile command-line is explicitly asking for librapidsml_jni.so to be dynamically linked against Arrow:
```
-Wl,-Bdynamic  /root/miniconda3/lib/libcudf.so  /usr/local/cuda/lib64/libcublas.so  /root/miniconda3/lib/libarrow.so.500.0.0  /root/miniconda3/lib/libarrow_cuda.so.500.0.0
```

I suspect the linker on Ubuntu is defaulting to omitting shared libs if the symbols in the libraries are not needed, whereas the CentOS 7 linker is always including the shared lib even if it is not referenced by symbols.

jlowe:
You could try modifying the link line to add `-Wl,--as-needed` before the `-Wl,-Bdynamic` to see if it changes the linking behavior on CentOS 7 to no longer include the Arrow libs.

wjxiz1992:
closed by https://github.com/NVIDIA/spark-rapids-ml/pull/39

PR #40, opened 2021-12-14
Move SVD calculation from Driver to Executor

wjxiz1992:
To fix https://github.com/NVIDIA/spark-rapids-ml/issues/41

### DIFF
--- a/src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
+++ b/src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
@@ -88,24 +88,23 @@ class RapidsRowMatrix(
     if (useCuSolverSVD) {
       val nvtxRangeSVD = new NvtxRange("cuSolver SVD", NvtxColor.BLUE)
 
-      val dense_U = DenseMatrix.zeros(n, n)
 
-      val dense_S = DenseMatrix.zeros(1, n)
-      var svdDriver: Array[(DenseMatrix, DenseMatrix)] = null
+      var svdOutput: Array[(DenseMatrix, DenseMatrix)] = null
       try {
-        val rdd_tmp = rows.context.parallelize(Seq(0),1)
-        val svdRdd = rdd_tmp.mapPartitions( p => {
+        val rddTmp = rows.context.parallelize(Seq(0), 1)
+        val dense_U = DenseMatrix.zeros(n, n)
+
+        val dense_S = DenseMatrix.zeros(1, n)
+        val svdRdd = rddTmp.mapPartitions( _ => {
           RAPIDSML.calSVD(n, Cov, dense_U, dense_S, 0)
           Iterator.single(dense_U, dense_S)
         })
-        svdDriver = svdRdd.collect()
+        svdOutput = svdRdd.collect()
       } finally {
         nvtxRangeSVD.close()
       }
-      val dense_U_new = svdDriver.head._1
-      val dense_S_new = svdDriver.head._2
-      val u = dense_U_new.asBreeze.asInstanceOf[BDM[Double]]
-      val s = dense_S_new.asBreeze.asInstanceOf[BDM[Double]]
+      val u = svdOutput.head._1.asBreeze.asInstanceOf[BDM[Double]]
+      val s = svdOutput.head._2.asBreeze.asInstanceOf[BDM[Double]]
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
```diff
@@ -91,14 +91,21 @@ class RapidsRowMatrix(
       val dense_U = DenseMatrix.zeros(n, n)
 
       val dense_S = DenseMatrix.zeros(1, n)
+      var svdDriver: Array[(DenseMatrix, DenseMatrix)] = null
       try {
-        // this is done on driver, so no task resources here, assign 0 manually.
-        RAPIDSML.calSVD(n, Cov, dense_U, dense_S, 0)
+        val rdd_tmp = rows.context.parallelize(Seq(0),1)
```
wbo4958, line 96: There should be an space after ","

and better to follow the java naming conventions https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
```diff
@@ -91,14 +91,21 @@ class RapidsRowMatrix(
...
+        val svdRdd = rdd_tmp.mapPartitions( p => {
+          RAPIDSML.calSVD(n, Cov, dense_U, dense_S, 0)
```
wbo4958, line 98: The map function needs to serialize the variables it needs and send them to the executor. So you may need to consider if you need to initialize them (like dense_U, dense_S) in the executor directly.

wbo4958 (commented):
Some minor comments.

src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
wjxiz1992: Thx, optimized.

wbo4958: Looks like you have specified the `device_id = 0` here, which may cause some issues when the worker has 2 GPUS. and the task is scheduled to GPU device=1.

wjxiz1992: nice catch. Fixed.

src/main/scala/org/apache/spark/ml/linalg/distributed/RapidsRowMatrix.scala
```diff
@@ -88,17 +88,29 @@ class RapidsRowMatrix(
...
+      var svdOutput: Array[(DenseMatrix, DenseMatrix)] = null
+      val gpuIdBC = rows.context.broadcast(gpuId)
       try {
-        // this is done on driver, so no task resources here, assign 0 manually.
-        RAPIDSML.calSVD(n, Cov, dense_U, dense_S, 0)
+        val rddTmp = rows.context.parallelize(Seq(0), 1)
+
+        val svdRdd = rddTmp.mapPartitions( _ => {
+          val gpu = if (gpuIdBC.value == -1) {
+            TaskContext.get().resources()("gpu").addresses(0).toInt
```
wbo4958, line 99: This will throw exceptions in local mode.

Try

  def getGPUAddrFromResources: Int = {
    val tc = TaskContext.get()
    if (tc == null) {
      throw new RuntimeException("Something wrong for task context")
    }
    val resources = tc.resources()
    if (resources.contains("gpu")) {
      val addrs = resources("gpu").addresses
      if (addrs.size > 1) {
        // TODO should we throw exception ?
        logger.warn("XGBoost only supports 1 gpu per worker")
      }
      // take the first one
      addrs.head.toInt
    } else {
      throw new RuntimeException("gpu is not allocated by spark, " +
        "please check if gpu scheduling is enabled")
    }
  }

wjxiz1992: We have a dedicated config for this case `PCA().setGpuId()` when we see the exception like "resource no key for gpu".
It exists on both training and transform. I think I can apply this fix for it in another follow-up PR for better UX.

NvTimLiu (approved):

PR #50, opened 2021-12-20
Update README

wjxiz1992:
Update description for the API usage. Now the API is exact the same as CPU. Only different thing is setInputCol is accepting column of ArrayType, not Vector type.


rongou (approved):

sameerz:
Looks ok to me, I am not familiar with the API so not going to sign off one way or the other.

Issue #57, opened 2022-02-01
Crash when calling fit passing a DataFrame


razajafri:
I am in the process of bechmarking different runtimes for using this library but when I use a file that is 1.6 gb, the spark job crashes with the following exception 

```
22/01/31 21:02:31 WARN package: Truncated the string representation of a plan since it was too large. This behavior can be adjusted by setting 'spark.sql.debug.maxToStringFields'.
22/01/31 21:02:32 WARN DAGScheduler: Broadcasting large task binary with size 1105.3 KiB
22/01/31 21:02:50 WARN DAGScheduler: Broadcasting large task binary with size 1970.1 KiB
22/01/31 21:04:00 WARN TaskSetManager: Lost task 12.0 in stage 5.0 (TID 41) (192.168.1.250 executor 0): java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:430)
	at scala.collection.immutable.Nil$.head(List.scala:427)
	at org.apache.spark.ml.linalg.distributed.RapidsRowMatrix.$anonfun$computeCovariance$1(RapidsRowMatrix.scala:134)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitions$2(RDD.scala:863)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitions$2$adapted(RDD.scala:863)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

The script being used to write the parquet file is 

```
[...]

PR #58, opened 2022-02-08
crash fix for empty partition

razajafri:
There were a couple of things that I thought were the reason for the crash and also for scaling with different partitions 

1. The code crashed when it received an empty partition. This was happening on larger files 
2. I was getting OOM Exception because the partition wasn't being closed. 

I have tried my best to write this but I don't understand the PCA algorithm and what it's supposed to do. I don't know if this is the right way to fix this.


wjxiz1992:
Thanks a lot Raza! I am also confused why there's an empty partition when larger file comes in. Have you experienced such case?

wjxiz1992:
Hi Raza, I posted another fix that uses mapPartitions method https://github.com/NVIDIA/spark-rapids-ml/pull/59. I haven't tested the map method one, is there a big perf difference between map vs. mapPartitions?

wjxiz1992:
During the mapPartitions process, I print the length for each partition and find in most cases, there's only one table in a partition with default config. So I guess there's no noticeable difference between map and mapPartitions. But let me do more tests with smaller partitions.

wjxiz1992 (commented):
I tested this change with my 4Gb parquet file, there's no crash or performance drop compared to the mapPartitions way.

razajafri:
> Thanks a lot Raza! I am also confused why there's an empty partition when larger file comes in. Have you experienced such case?

Yes, I posted a bug with the stack trace in it [here](https://github.com/NVIDIA/spark-rapids-ml/issues/57)

> Hi Raza, I posted another fix that uses mapPartitions method https://github.com/NVIDIA/spark-rapids-ml/pull/59. I haven't tested the map method one, is there a big perf difference between map vs. mapPartitions?

I haven't benchmarked the `mapPartitions` because it was crashing. The way you have written the mapPartitions concatenates all the tables into a single large table, which is unnecessary as the caller can do the processing over each table the same when it iterates over them.

rongou (approved):

PR #61, opened 2022-02-17
Change version and dependency jars to 22.04.0-SNAPSHOT

NvTimLiu:
Signed-off-by: Tim Liu <timl@nvidia.com>


wjxiz1992 (approved):
LGTM, we also need to update the example code in the spark-rapids-examples repo.

PR #71, opened 2022-05-30
remove cuDF jar dependency

wjxiz1992:

After https://github.com/NVIDIA/spark-rapids/pull/5249, cuDF dependency has been intergrated into spark-rapids jar. There's no need to use cuDF denpendency in the pom file, so we remove it in this PR.

This change has been tested with the [example app](https://github.com/NVIDIA/spark-rapids-examples/tree/branch-22.06/examples/ML%2BDL-Examples/Spark-cuML/pca)

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,20 +73,19 @@ the way. Make sure the prerequisites are all met, or the build will fail with er
 accordingly such as "cmake not found" or "ninja not found" etc.
 
 ## How to use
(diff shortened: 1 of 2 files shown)


pxLi:
can we also include the doc change here? to update https://github.com/NVIDIA/spark-rapids-ml#how-to-use

pxLi (approved):
thx

PR #108, opened 2023-02-03
fix raft version to 22.12

wjxiz1992:
Since https://github.com/rapidsai/raft/commit/79e1ce854329bd655f2b02c86e6ec39165eebfc6, raft changes their API again and this breaks our build pipeline. Since we don't plan to introduce any new functionalities to this repo, we fix the raft version to 22.12 to avoid such issue in the future.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -90,7 +90,7 @@ RUN wget --quiet \
 ARG CONDA_CUDF_VER=22.04
 RUN conda install -c rapidsai -c conda-forge cudf=$CONDA_CUDF_VER python=3.8 -y
 
-ARG RAFT_VER=23.02
+ARG RAFT_VER=22.12
 RUN git clone -b branch-$RAFT_VER https://github.com/rapidsai/raft.git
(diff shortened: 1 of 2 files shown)


wjxiz1992:
@pxLi FYI, we now fix the raft version to 22.12 so we can just skip modifying raft version value when we want to release a new version of spark-rapids-ml.

docker/Dockerfile
```diff
@@ -90,7 +90,7 @@ RUN wget --quiet \
 ARG CONDA_CUDF_VER=22.04
 RUN conda install -c rapidsai -c conda-forge cudf=$CONDA_CUDF_VER python=3.8 -y
 
-ARG RAFT_VER=23.02
+ARG RAFT_VER=22.12
```
GaryShen2008, line 93: Can we add some comments to mention we fixed the version?

wjxiz1992: Sure, updated.

GaryShen2008 (approved):

PR #118, opened 2023-02-24
clarify db instructions and top level readme

eordentlich:


notebooks/databricks/README.md
```diff
@@ -1,8 +1,42 @@
...
-- Create a cluster using Databricks 10.4LTS runtime (11.3 is not yet compatible with spark-rapids) using at least 2 gpu based workers and add the init file dbfs location used above to the spark-config section of the cluster and using the Spark configs below:
-```
-```
-- Select your workspace and upload desired notebook via `Import` in the drop down menu for your workspace.
+## Running notebooks on Databricks
+Assuming you already have a Databricks account, to run notebooks on Databricks do the following:
+- If you don't already have it, install [databricks cli](https://learn.microsoft.com/en-us/azure/databricks/dev-tools/cli/) and create and save an [access token](https://docs.databricks.com/dev-tools/api/latest/authentication.html) to your workspace using the workspace UI.  You may need to create a new profile, in which case it should be supplied via the `--profile` option to all `databricks` cli commands issued below.
```
leewyang, line 3: Maybe use this link for databricks-cli: https://docs.databricks.com/dev-tools/cli/index.html

leewyang (approved):

notebooks/databricks/README.md
eordentlich: done

leewyang (approved):

PR #124, opened 2023-02-28
refactor benchmarks

leewyang:
- moves linear regression and random forest benchmarks to separate files.
- standardizes on `BenchmarkRunner` and `BenchmarkBase`, with the following updates:
  - infers command line arguments from class under test, with separate params for Spark and cuML (or both).
  - adds standardized support for a reporting dictionary (w/ persistence).
  - adds support for multiple iterations.
- updates `./run_benchmarks.sh` to run either "all" tests, or a single test.
  - defaults to single-node, single-gpu confs for testing in dev and CI/CD envs.

Note: I tried to leave the core benchmarking code as similar as possible, so there is likely still room for more refactoring.  That said, this puts all of the existing benchmarks on the same base framework, so future cleanup should hopefully be easier and less disruptive.


eordentlich (commented):
Any chance `bench_*.py` can also be moved to benchmark subdirectory with import statements in benchmark_runner.py changed accordingly?  Makes it a bit easier to run these in databricks via the cli to have only one top level file.

benchmark/benchmark/bench_pca.py
```diff
@@ -0,0 +1,258 @@
...
+        total_score = total_score['total_score'][0]
+        return orthonormality_score, total_score
+
+    def run_once(
+        self,
+        spark: SparkSession,
+        df: DataFrame,
+        features_col: Union[str, List[str]],
+        label_name: Optional[str],
+    ) -> None:
```
lijinf2, line 106: Seems a typo. Use Dict[str, Any] as the return type instead of None?

benchmark/benchmark/bench_kmeans.py
```diff
@@ -0,0 +1,244 @@
...
+        total_score = total_score['total_score'][0]
+        return total_score
+
+    def run_once(
+        self,
+        spark: SparkSession,
+        df: DataFrame,
+        features_col: Union[str, List[str]],
+        label_name: Optional[str],
+    ) -> None:
```
lijinf2, line 105: Typo: Replace None with Dict[str, Any] as return type annotation?

lijinf2:
Looks Good!

benchmark/benchmark/bench_pca.py
leewyang: Good catch (turned out these files weren't being type checked until I moved them into the subfolder... so fixed that as well.

PR #125, opened 2023-02-28
[Draft PR] Add GPU-Accelerated NearestNeighbors on Spark

lijinf2:


src/spark_rapids_ml/core.py
```diff
@@ -438,15 +454,20 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
-            .collect()[0]
-        )
+        barrier_rdd = dataset.mapInPandas(
+            _train_udf, schema=self._out_schema()
+        ).rdd.barrier()  # type: ignore
+
+        if return_model == False:
+            return barrier_rdd.mapPartitions(lambda x: x)
+
+        ret = barrier_rdd.mapPartitions(lambda x: x).collect()[0]
```
leewyang, line 470: I'd move `mapPartitions(lambda x: x)` into the common code above, since it's duplicated otherwise.

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,344 @@
...
+    Shared Spark Params for NearestNeighbor and NearestNeighborModel.
+    """
+
+    def setInputCol(self, value: str) -> "_NearestNeighborsCumlParams":
+        """
+        Sets the value of `inputCol`.
+        """
+        self.set_params(inputCol=value)
+        return self
+
```
leewyang, line 83: Do we also need `setLabelCol()`?

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,344 @@
...
+        self.label_isquery = 1
+        self.row_number_col = alias.row_number
+        self.set_params(labelCol=alias.label)
+
+    def setK(self, value: int) -> "NearestNeighbors":
+        """
+        Sets the value of `k`.
+        """
+        self.set_params(n_neighbors=value)
+        return self
```
leewyang, line 145: This should be handled via parameter mapping, since mapping a Spark Param value to a differently named cuML param leads to weird behavior/inconsistencies.  So, add a `k` Spark ML Param and a `_param_mapping` entry to map Spark `k` to cuML `n_neighbors`.  (See `KMeans` for reference).

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,344 @@
...
+    """
+    Examples
+    --------
+    >>> from sparkcuml.neighbors import NearestNeighbors
+    >>> data = [([1.0, 1.0],),
+    ...         ([2.0, 2.0],),
+    ...         ([3.0, 3.0],),]
+    >>> topk = 2
+    >>> gpu_knn = NearestNeighbors().setInputCol("features").setK(topk)
+    >>> data_df = spark.SparkContext.parallelize(data).map(lambda row: (row,)).toDF(["features"])
```
leewyang, line 119: Is the map to tuple required here?  Doesn't seem required in unit tests.

leewyang (commented):
A couple comments

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,344 @@
...
+    def _get_cuml_transform_func(
+        self, dataset: DataFrame
+    ) -> Tuple[
+        Callable[..., CumlT],
+        Callable[[CumlT, Union[cudf.DataFrame, np.ndarray]], pd.DataFrame],
+    ]:
+        pass
+
+
+_set_pyspark_cuml_cls_param_attrs(NearestNeighbors, NearestNeighborsModel)
```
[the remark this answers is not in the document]
lijinf2: Yeah, removed the functional call and tested good.

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,344 @@
...
+            context = BarrierTaskContext.get()
+            rank = context.partitionId()
+
+            from cuml.neighbors.nearest_neighbors_mg import NearestNeighborsMG as cumlNN
+
+            nn_object = cumlNN(
+                handle=params["handle"],
+                n_neighbors=params[INIT_PARAMETERS_NAME]["n_neighbors"],
+                verbose=params[INIT_PARAMETERS_NAME]["verbose"],
+            )
```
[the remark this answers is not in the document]
lijinf2: Added MG class into the list of _cuml_cls. 
Edited the batch_size parameter into the test case test_nearest_neighbors.

src/spark_rapids_ml/core.py
```diff
@@ -358,7 +360,13 @@ def _enable_nccl(self) -> bool:
...
         return True
 
-    def _fit(self, dataset: DataFrame) -> "_CumlModel":
+    def _require_ucx(self) -> bool:
+        """If enable or disable ucx over NCCL"""
+        return False
+
+    def _fit(
+        self, dataset: DataFrame, return_model: bool = True
+    ) -> Union["_CumlModel", RDD]:
```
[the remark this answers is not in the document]
lijinf2: Added a return_model function in the similar way as enable_nccl and enable_ucx.

src/spark_rapids_ml/core.py
lijinf2: Done.

src/spark_rapids_ml/knn.py
lijinf2: Ah..., probably not. NN reused the labelCol to differentiate data and query vectors, but the standard NN application itself does not need the label column. I may be wrong.

src/spark_rapids_ml/knn.py
lijinf2: done.

src/spark_rapids_ml/knn.py
lijinf2: done.

leewyang (approved):
LGTM

eordentlich (commented):
In this example: https://spark.apache.org/docs/3.2.1/api/python/reference/api/pyspark.ml.feature.BucketedRandomProjectionLSH.html
there is an 'id' column.    We should probably support that.  

I'm actually not even sure that 'order' of the starting DF -> rdd is deterministic necessarily now that I think about it.

lijinf2:
> In this example: https://spark.apache.org/docs/3.2.1/api/python/reference/api/pyspark.ml.feature.BucketedRandomProjectionLSH.html there is an 'id' column. We should probably support that.
> 
> I'm actually not even sure that 'order' of the starting DF -> rdd is deterministic necessarily now that I think about it.

Marked and will do (1) support id column (or neighboring columns) (2) ucx asyncio fix. 
Seem non-trivial so perhaps move to the next PRs to avoid this PR being too long? If so, will need approval again for this PR.

src/spark_rapids_ml/core.py
```diff
@@ -438,16 +457,23 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
             if enable_nccl:
                 context.barrier()
 
-            if context.partitionId() == 0:
+            if return_model == True:
+                if context.partitionId() == 0:
+                    yield pd.DataFrame(data=result)
+            else:
```
wbo4958, line 463: looks like this doesn't require context.barrier in line  439?

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+
+            distances = [ary.tolist() for ary in distances]
+            indices = [ary.tolist() for ary in indices]
+            query = [ary.tolist() for ary in query]  # type: ignore
+            item = [ary.tolist() for ary in item]  # type: ignore
+
+            # id mapping
+            id2row: Dict[int, int] = {}
+            count = 0
+            for r, _, item_rn in triplets:
```
eordentlich, line 313: Lot's of python for looping here and below (though some comprehension too).  We'll need to check how this does on GPU size scale data relative to the cuml part.

tests/test_nearest_neighbors.py
```diff
@@ -0,0 +1,144 @@
...
+        assert indices[2] == [3, 4]
+
+        assert array_equal(distances[3], [0.0, math.sqrt(2.0)])
+        assert indices[3] == [7, 6]
+
+        assert array_equal(distances[4], [math.sqrt(2.0), math.sqrt(8.0)])
+        assert indices[4] == [7, 6]
+
+
+@pytest.mark.parametrize("feature_type", ["array"])
```
eordentlich, line 73: Does this mean only 'array' input type is currently supported?

eordentlich:
The 

> > In this example: https://spark.apache.org/docs/3.2.1/api/python/reference/api/pyspark.ml.feature.BucketedRandomProjectionLSH.html there is an 'id' column. We should probably support that.
> > I'm actually not even sure that 'order' of the starting DF -> rdd is deterministic necessarily now that I think about it.
> 
> Marked and will do (1) support id column (or neighboring columns) (2) ucx asyncio fix. Seem non-trivial so perhaps move to the next PRs to avoid this PR being too long? If so, will need approval again for this PR.

For now let's at least return the input df's with the added zipped indices so that all the info is there and it is technically correct, though some post processing might be needed.   We can follow up and add a exact knn join api similar to the approxSimilarity join api of SparkML which lists all pairs of rows in dfA and dfB that are knn neighbors with the distance.

src/spark_rapids_ml/core.py
```diff
@@ -358,7 +360,15 @@ def _enable_nccl(self) -> bool:
...
-    def _fit(self, dataset: DataFrame) -> "_CumlModel":
+    def _require_ucx(self) -> bool:
+        """If enable or disable ucx over NCCL"""
+        return False
+
+    def _return_model(self) -> bool:
+        """If _fit returns a model or a RDD"""
+        return True
+
+    def _fit(self, dataset: DataFrame) -> Union["_CumlModel", RDD]:
```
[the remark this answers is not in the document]
wbo4958: if you really want to return RDD. I'd like to suggest you move the code to a separate function to reuse some code instead of changing the "_fit"

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def _create_pyspark_model(self, result: Row) -> "NearestNeighborsModel":
+        return NearestNeighborsModel.from_row(result)
+
+    def fit(self, dataset: DataFrame, params: Optional[Dict[Param[Any], Any]] = None) -> "NearestNeighbors":  # type: ignore
+        self.data_df = dataset.withColumn(alias.label, lit(self.label_isdata))
+        self.data_df = self.df_zip_with_index(self.data_df)
+        return self
+
+    def kneighbors(self, query_df: DataFrame) -> Tuple[DataFrame, DataFrame]:
```
wbo4958, line 171: seems this is not pyspark ML way, the method neighbors should be in KNN model?

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+        res_df = res_rdd.toDF(
+            schema="query_index int, indices array<int>, distances array<float>"
+        )
+        res_df = res_df.sort("query_index")
+        distances_df = res_df.select("distances")
+        indices_df = res_df.select("indices")
+        return (distances_df, indices_df)
+
+    def df_zip_with_index(self, df: DataFrame) -> DataFrame:
```
wbo4958, line 190: Could you add some comments on this function to describe what's this API doing?

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+        out_schema = StructType(
+            [StructField(self.row_number_col, IntegerType(), False)] + df.schema.fields
+        )
+        zipped_rdd = df.rdd.zipWithIndex()
+        new_rdd = zipped_rdd.map(lambda row: [row[1]] + list(row[0]))
+        return new_rdd.toDF(schema=out_schema)
```
wbo4958, line 196: according to https://stackoverflow.com/a/56651510/1928940, this solution may be better, especially for spark-rapids.

eordentlich: As written, does the added column actually need to be contiguous?  A mapping is constructed internally to the code and it doesn't assume anything about this column.

src/spark_rapids_ml/knn.py
eordentlich: Isn't internal.  Probably should be private with `_` prefix.

src/spark_rapids_ml/knn.py
eordentlich: Same goes for the Dict lookups.

src/spark_rapids_ml/common/cuml_context.py
```diff
@@ -13,57 +13,142 @@
...
+        else:
+            tasks = context.getTaskInfos()
+            ips = [task.address.split(":")[0] for task in tasks]
+
+            # set environmental variables according to https://github.com/rapidsai/ucx-py
+            # the code occasionally run fail without setting the variables
+            my_ip = ips[self._rank]
+            my_ifname = CumlContext.get_ifname_from_ip(my_ip)
+            os.environ["UCX_TLS"] = "tcp,cuda_copy,cuda_ipc"
+            os.environ["UCXPY_IFNAME"] = my_ifname
```
eordentlich, line 82: We'll have to figure how to make this more flexible to take advantage of higher speed interconnects on multi-gpu nodes.

src/spark_rapids_ml/common/cuml_context.py
```diff
@@ -13,57 +13,142 @@
...
+    ) -> Endpoint:
+        """
+        ucp initialization may require a larger additional_timeout a complex network environment
+        """
+        eps = [None] * len(target_ip_ports)
+        for i in range(len(eps)):
+            ip, port = target_ip_ports[i]
+            ep = await ucx_worker.get_endpoint(ip, port)
+            eps[i] = ep
+        await asyncio.sleep(additional_timeout)
```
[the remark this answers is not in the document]
lijinf2: created a ticket.

src/spark_rapids_ml/core.py
lijinf2: revised.

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+        res_rdd = pipelinedrdd.flatMap(
+            lambda row: list(zip(row["query_index"], row["indices"], row["distances"]))
+        )
+        res_df = res_rdd.toDF(
+            schema="query_index int, indices array<int>, distances array<float>"
+        )
+        res_df = res_df.sort("query_index")
+        distances_df = res_df.select("distances")
+        indices_df = res_df.select("indices")
+        return (distances_df, indices_df)
```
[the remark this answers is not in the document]
lijinf2: Revised code to return (query_df, item_df ,knn_df), where knn_df has columns (query_id, knn_indices, knn_distances).

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+            item_query_sizes = [len(chunk) for chunk in item] + [
+                len(chunk) for chunk in query
+            ]
+
+            import json
+
+            messages = context.allGather(
+                message=json.dumps((rank, item_query_sizes, item_row_number))
+            )
+            triplets = [json.loads(msg) for msg in messages]
```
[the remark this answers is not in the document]
lijinf2: revised the variable name.

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+
+            item_parts_to_ranks = []
+            query_parts_to_ranks = []
+            for r, sizes, _ in triplets:
+                half_len = len(sizes) // 2
+                item_parts_to_ranks += [(r, size) for size in sizes[:half_len]]
```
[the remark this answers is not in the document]
lijinf2: revised.

src/spark_rapids_ml/knn.py
lijinf2: created a ticket for future improvement.

tests/test_nearest_neighbors.py
lijinf2: Yeah, created a ticket to support multi-col df and vector df.

src/spark_rapids_ml/core.py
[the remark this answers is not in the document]
lijinf2: Focusing on KNN functionality for now. 

Created a ticket to improve coding style in the future.

src/spark_rapids_ml/knn.py
lijinf2: copied this comment to the last comment and the created ticket.

src/spark_rapids_ml/knn.py
lijinf2: Revised.

src/spark_rapids_ml/knn.py
lijinf2: Not sure about the behavior of monotonically_increasing_id. Added a TODO comment and created a ticket.

src/spark_rapids_ml/common/cuml_context.py
lijinf2: created a ticket.

lijinf2:
@eordentlich @wbo4958 Addressed some comments and created tickets for the rest that seem to require further investigation. 
Please help review the PR again. Thanks.

src/spark_rapids_ml/knn.py
```diff
@@ -90,15 +90,42 @@ class _NearestNeighborsCumlParams(_CumlParams, HasInputCol, HasLabelCol):
...
     def setInputCol(self, value: str) -> "_NearestNeighborsCumlParams":
         """
         Sets the value of `inputCol`.
         """
         self.set_params(inputCol=value)
         return self
 
+    def setIdCol(self, value: str) -> "_NearestNeighborsCumlParams":
+        """
+        Sets the value of `id_col`. If not set, zipWithIndex will be used and the row order is not guaranted to be deterministic.
```
eordentlich, line 109: Is row order ever guaranteed to be deterministic? So not sure what is meant here.  Should just say an id column will be added with column name [...]  no need to mention zipWithIndex.   Also, somewhere should explain what an id column is used for (e.g., to specify nearest neighbor vectors by associated id value).

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,400 @@
...
+        return select_cols, multi_col_names, dimension, feature_type
+
+
+class NearestNeighbors(
+    NearestNeighborsClass, _CumlEstimatorNearestNeighbors, _NearestNeighborsCumlParams
+):
+    """
+    Examples
+    --------
+    >>> from spark_rapids_ml.knn import NearestNeighbors
```
eordentlich, line 160: The example seems to need updating to match latest changes.

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,400 @@
...
+    ]:
+        (
+            select_cols,
+            multi_col_names,
+            dimension,
+            feature_type,
+        ) = super()._pre_process_data(dataset)
+
+        if self.hasParam("id_col") and self.isDefined("id_col"):
+            label_name = self.getOrDefault("id_col")
```
eordentlich, line 146: Why label_name as variable name here?

src/spark_rapids_ml/knn.py
```diff
@@ -0,0 +1,362 @@
...
+            query_row_number = [query_row_number]
+
+            item_query_sizes = [len(chunk) for chunk in item] + [
+                len(chunk) for chunk in query
+            ]
+
+            import json
+
+            messages = context.allGather(
+                message=json.dumps((rank, item_query_sizes, item_row_number))
```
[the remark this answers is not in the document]
eordentlich: Any comment on this?

src/spark_rapids_ml/knn.py
lijinf2: Revised the comment. Please check.

src/spark_rapids_ml/knn.py
lijinf2: Good catch. Revised the variable name to id_col_name.

src/spark_rapids_ml/knn.py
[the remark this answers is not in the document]
lijinf2: Good question. Created a ticket to investigate the performance of json dumps/loads and the potential to leverage ucx.

eordentlich:
Can you rebase?

lijinf2:
> Can you rebase?

Done. Please check.

src/spark_rapids_ml/core.py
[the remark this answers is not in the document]
wbo4958: Could you move the common code in "_fit" into another function (eg, abc_func) to return the RDD. and in KNN please overwrite fit() which should call your abc_func instead of "_fit"

eordentlich:
👍

eordentlich (approved):

src/spark_rapids_ml/core.py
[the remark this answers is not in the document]
lijinf2: Copied the latest comment into the ticket for record. Will do this in the next PRs to avoid this one being too long.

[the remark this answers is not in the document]
eordentlich: @wbo4958 @lijinf2 will do this or similar  in a followup pr.

PR #129, opened 2023-03-01
fix benchmark args

leewyang:
This reverts some of the changes introduced in #124 to fix:
- Spark 3.2.1 (which doesn't have type annotations for TypeConverters)
- CPU testing w/o GPU dependencies


eordentlich (approved):
👍

PR #132, opened 2023-03-07
RF: map to the correct max_features

wbo4958:
This PR maps the max_features to the correct value when it is "auto",

Basically follow below rules,

``` console
   *   - "auto": Choose automatically for task:
   *            If numTrees == 1, set to "all."
   *            If numTrees greater than 1 (forest), set to "sqrt" for classification and
   *              to "onethird" for regression.
```

### DIFF
--- a/src/spark_rapids_ml/tree.py
+++ b/src/spark_rapids_ml/tree.py
@@ -87,7 +87,7 @@ class _RandomForestClass(_CumlClass):
     def _param_value_mapping(cls) -> Dict[str, Dict[str, Union[str, None]]]:
         return {
             "max_features": {
-                "onethird": "0.3333",
+                "onethird": str(1 / 3.0),
                 "all": "1.0",
                 "auto": "auto",
                 "sqrt": "sqrt",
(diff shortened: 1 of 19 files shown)


src/spark_rapids_ml/tree.py
```diff
@@ -198,6 +200,12 @@ def _rf_fit(
             rf_params = params[INIT_PARAMETERS_NAME]
             rf_params.pop("n_estimators")
 
+            if rf_params["max_features"] == "auto":
+                if total_trees == 1:
+                    rf_params["max_features"] = 1.0
+                else:
+                    rf_params["max_features"] = "sqrt" if is_classification else 0.33333
```
lijinf2, line 207: PySpark says 'onethird' in the doc of method trainRegressor(). Is it equivalent to 0.33333 in the implementation?

eordentlich: Very close.   Just put 1.0/3.0

wbo4958: Done. Thx

leewyang (approved):

eordentlich (approved):
👍

PR #133, opened 2023-03-07
[NearestNeighbors] add support for two input dataframe types, multi_cols and VectorUDT

lijinf2:
To support multi_cols and VectorUDT input format. Tests added.


eordentlich (approved):
👍

PR #136, opened 2023-03-09
dockerfiles for python

leewyang:
- `Dockerfile.python` - uses conda and cuda 11.5 (based on existing Scala dockerfile)
- `Dockerfile.pip` -uses pip and cuda 11.8 (based on Erik's init scripts for Databricks)

### DIFF
--- a/docker/Dockerfile.pip
+++ b/docker/Dockerfile.pip
@@ -1,0 +1,58 @@
+#
+# Copyright (c) 2023, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+ARG CUDA_VERSION=11.8.0
+FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu20.04
+
+ARG PYSPARK_VERSION=3.3.1
+ARG RAPIDS_VERSION=22.12
+
+# Install packages to build spark-rapids-ml
+RUN apt-get update -y \
+    && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt install -y openjdk-8-jdk \
+    && rm -rf /var/lib/apt/lists
+
+RUN apt-get update -y \
+    && apt install -y git numactl python3.8 python3-pip software-properties-common wget zip \
+    && python3 -m pip install --upgrade pip \
+    && rm -rf /var/lib/apt/lists
+
+# install RAPIDS
+RUN pip install --no-cache-dir \
+    cudf-cu11==${RAPIDS_VERSION} \
+    cuml-cu11==${RAPIDS_VERSION} \
+    dask-cudf-cu11==${RAPIDS_VERSION} \
+    pylibraft-cu11==${RAPIDS_VERSION} \
+    raft-dask-cu11==${RAPIDS_VERSION} \
+    rmm-cu11==${RAPIDS_VERSION} \
+    --extra-index-url=https://pypi.nvidia.com
+
+# install pyspark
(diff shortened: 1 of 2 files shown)


lijinf2:
Looks good!

docker/Dockerfile.pip
```diff
@@ -0,0 +1,58 @@
...
+#ARG RAPIDS_ML_VER=spark-cuml
+#RUN git clone -b branch-$RAPIDS_ML_VER https://github.com/NVIDIA/spark-rapids-ml.git
+COPY . /spark-rapids-ml
+WORKDIR /spark-rapids-ml
+
+# install spark-rapids-ml
+RUN pip install --no-cache -r requirements_dev.txt \
+    && pip install --no-cache -e .
+
+SHELL ["conda", "run", "--no-capture-output", "-n", "base", "/bin/bash", "-c"]
```
eordentlich, line 58: Do we still need Conda here?

docker/Dockerfile.python
```diff
@@ -0,0 +1,108 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+###
```
eordentlich, line 18: Why retain all the commented out commands?

docker/Dockerfile.python
```diff
@@ -0,0 +1,108 @@
...
+#        --build-arg CUDA_VERSION=11.5.2 \
+#        --build-arg GCC_VERSION=9 \
+#        --build-arg CMAKE_VERSION=3.23.3 \
+#        --build-arg NINJA_VERSION=1.10.2 \
+#        --build-arg MAVEN_VERSION=3.8.6
+#
+###
+
+ARG CUDA_VERSION=11.5.2
+FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu20.04
```
eordentlich, line 39: Same comment about devel base as for other file.

docker/Dockerfile.pip
```diff
@@ -0,0 +1,58 @@
...
+
+# install pyspark
+RUN pip install --no-cache-dir pyspark==${PYSPARK_VERSION}
+
+# Config JAVA_HOME
+ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk-amd64
+
+#ARG RAPIDS_ML_VER=spark-cuml
+#RUN git clone -b branch-$RAPIDS_ML_VER https://github.com/NVIDIA/spark-rapids-ml.git
+COPY . /spark-rapids-ml
```
eordentlich, line 51: Are we sure docker build will be run in src directory?

eordentlich (commented):
We should alsoadd instructions for use to the docker dir README.md
https://github.com/NVIDIA/spark-rapids-ml/blob/spark-cuml/docker/README.md

docker/Dockerfile.pip
```diff
@@ -0,0 +1,58 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+ARG CUDA_VERSION=11.8.0
+FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu20.04
```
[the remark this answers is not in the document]
leewyang: Original docker image used `devel`, so I just kept it.  Just tried using a `runtime` image, but got errors during tests, so we can revisit slimming the image later.

docker/Dockerfile.pip
leewyang: Good catch... removed.

docker/Dockerfile.python
leewyang: Vestige of original dockerfile... removed.

docker/Dockerfile.python
leewyang: See reply above.

docker/Dockerfile.pip
leewyang: This is actually copying the entire repo (at the root-level) into the docker image, so the CI commands would look like:
```
git clone git@github.com:NVIDIA/spark-rapids-ml.git
cd spark-rapids-ml
docker build -t <image_tag> -f docker/Dockerfile .
```

eordentlich (approved):
👍   Should follow up with addition to README

PR #135, opened 2023-03-08
avoid ucx manual sleep with an automatic method using lower-level asy…

lijinf2:
Copied and merged ucx code snippets from [gitlab branch](https://gitlab-master.nvidia.com/eordentlich/spark-cuml/-/blob/a43951b5cfd1cf99bb50c35b23ed00fa7a0e4e56/src/spark_rapids_ml/common/cuml_context.py)

Added tests for UCX. 
Tests got passed in both workstation and dgx.2.

PR #137, opened 2023-03-09
Fix a RF test bug

wbo4958:
Fix two bugs introduced by #135 and https://github.com/NVIDIA/spark-rapids-ml/pull/132

### DIFF
--- a/src/spark_rapids_ml/common/cuml_context.py
+++ b/src/spark_rapids_ml/common/cuml_context.py
@@ -50,6 +50,8 @@ class CumlContext:
         """
         self.enable = enable
         self._handle: Optional[Handle] = None
+        self._loop: Optional[AbstractEventLoop] = None
+
         if not enable:
             return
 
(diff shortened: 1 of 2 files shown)


src/spark_rapids_ml/common/cuml_context.py
```diff
@@ -50,6 +50,8 @@ def __init__(
         """
         self.enable = enable
         self._handle: Optional[Handle] = None
+        self._loop: Optional[AbstractEventLoop] = None
```
wbo4958, line 53: RandomForest doesn't enable NCCL, but the _loop has been referred in the common core

eordentlich: Did the unit tests not catch this?

leewyang: Unit tests currently fail

leewyang (approved):

PR #141, opened 2023-03-10
fix RandomForestRegressor compat tests

leewyang:
Addresses outstanding TODO items to investigate differences in prediction results GPU and CPU variants, which looks related to the `num_workers` field.  So just forcing this to 1 during the compat tests.


eordentlich (approved):
👍

PR #142, opened 2023-03-10
expand docsite

leewyang:
... with some high-level sections, and pushed python api docs down a level to `api/python`.


eordentlich (approved):
👍

PR #144, opened 2023-03-11
fix benchmarks for py3.8

leewyang:
Fixes discrepancies in behavior of `inspect.signature()` between py3.8 and py3.9 by forcing the code to inspect the `__init__` function.

### DIFF
--- a/benchmark/benchmark/bench_kmeans.py
+++ b/benchmark/benchmark/bench_kmeans.py
@@ -33,7 +33,7 @@ class BenchmarkKMeans(BenchmarkBase):
         from pyspark.ml.clustering import KMeans
 
         params = inspect_default_params_from_func(
-            KMeans,
+            KMeans.__init__,
             [
                 "distanceMeasure",
                 "featuresCol",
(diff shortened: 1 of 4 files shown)


eordentlich:
How does this problem surface?   I haven't encountered any issues in a 3.8 env in DB.

leewyang:
Saw this behavior in a local conda env and in docker containers built from the dockerfiles.  When running `./run_benchmark.sh`, I got this error:
```
benchmark_runner.py: error: unrecognized arguments: --k 3
```

lijinf2:
The problem occurs when calling benchmark.utils.inspect_default_params_from_func(). Would the same problem exists in calling spark_rapids_ml.utils._get_default_params_from_func()? The two functions look very similar that have the same set of arguments (func: Callable, unsupported_set: List[str] = []).

eordentlich (approved):
👍

leewyang:
> The problem occurs when calling benchmark.utils.inspect_default_params_from_func(). Would the same problem exists in calling spark_rapids_ml.utils._get_default_params_from_func()? The two functions look very similar that have the same set of arguments (func: Callable, unsupported_set: List[str] = []).

Yeah, the problem seems to be isolated to the Spark class constructors.  The cuML constructors worked fine, so it looks like some combination of py3.8 + spark.

leewyang:
FYI, confirmed that the problem is specific to py3.8 + pyspark 3.3.  Pyspark 3.2 works fine with py3.8.

PR #145, opened 2023-03-13
data_gen supports train_fraction

wbo4958:
This PR adds `--train_size` for data_gen.py. With train_size, the train data will be wrote into output_dir/train and the eval data will be wrote into output_dir/eval

### DIFF
--- a/benchmark/gen_data.py
+++ b/benchmark/gen_data.py
@@ -123,7 +123,6 @@ class DataGenBase(DataGen):
             type=_restrict_train_size,  # type: ignore
             help="the value should be between 0.0 and 1.0 and represent "
             "the proportion of the dataset to include in the train split",
-            default=0.8,
         )
 
         self._add_extra_parameters()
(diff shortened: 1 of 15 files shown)


benchmark/gen_data.py
```diff
@@ -106,6 +106,26 @@ def __init__(self) -> None:
...
+            except ValueError:
+                raise argparse.ArgumentTypeError(f"{x} is not a floating-point literal")
+
+            if x < 0.0 or x > 1.0:
+                raise argparse.ArgumentTypeError(f"{x} is not in range [0.0, 1.0]")
+
+            return x
+
+        self._parser.add_argument(
+            "--train_size",
```
eordentlich, line 122: train_size -> train_fraction

eordentlich (commented):
looks good.  minor comment.

benchmark/gen_data.py
```diff
@@ -456,15 +476,23 @@ def make_classification_udf(iter: Iterator[pd.Series]) -> pd.DataFrame:
...
+        train_writer = train_df.write
+        eval_writer = eval_df.write
         if args.overwrite:
-            writer = df.write.mode("overwrite")
+            train_writer = train_df.write.mode("overwrite")
+            eval_writer = eval_df.write.mode("overwrite")
 
-        writer.parquet(args.output_dir)
+        train_writer.parquet(f"{args.output_dir}/train")
+        eval_writer.parquet(f"{args.output_dir}/eval")
```
leewyang, line 496: Can you also update `run_benchmark.sh` to use this new datagen (and update the various `--transform_path` args accordingly)?  (Could be in a followup PR)

lijinf2:
LGTM.

wbo4958:
I just pushed a new commit to unify the train and transform path,

eg,
``` python
python benchmark_runner.py random_forest_regressor --data_path=/tmp/xyz --transform
```

the train path will be set to /tmp/xyz/train and the transform path will be set /tmp/xyz/eval, and `--transform` indicates doing the evaluation.

wbo4958:
> @wbo4958 For the transform path, I think @eordentlich and I intended something more like this:
> 
> ```
> python benchmark_runner.py random_forest_regressor --train_path=/tmp/xyz/train --transform_path=/tmp/xyz/eval
> ```
> 
> ... where `datagen` generated both sets under a single folder, e.g. `/tmp/xyz/{train,eval}`, but we'd still want to allow independent paths if needed. So, essentially it'd just be a tweak to the `run_benchmark.sh` script to point to the newly generated `--train_path` and `--transform_path` locations. But in other environments, someone could still supply entirely separate paths, e.g. `--train_path /foo/train --transform_path /bar/baz/test`.

@leewyang Thx for the explanation, and one more question, if the user doesn't specify the transform_path, do we need to use the training dataset as the eval dataset? or just don't do the evaluation? I prefer the latter since it's really time-consuming to do the evaluation for the whole dataset for some algorithms if the data size is really large.

wbo4958:
For the transform_path parameter of benchmark runner, I'd like to have the following PR to handle it. So let's just focus to generate the train/eval dataset in this PR.

wbo4958:
> Can we retain current behavior as default if no train fraction arg supplied:
> 
> * no eval file produced.  train file has all specified rows.

Done, please help to check the latest commit

> * exactly equal number of rows in output part files if num rows is divisible by part files.

We have `--output_num_files` argument to control how many files will be generated. And it will introduce the repartition (RoundRobin manner), each part file will have the same row as long as num rows can be divisible by output_num_files

leewyang (approved):

PR #146, opened 2023-03-13
eliminate most 'F' order warnings

eordentlich:
in a memory efficient way and improve memory usage behavior in benchmarks.   

Also proactively deletes some objects.   

Found these to help with memory related reliability issues when running benchmarks.

One 'F' order warning persists, but I think it may be due to 1-d arrays being both 'C' and 'F' ordered and the check in cuml mnmg code not handling this correctly.

### DIFF
--- a/src/spark_rapids_ml/clustering.py
+++ b/src/spark_rapids_ml/clustering.py
@@ -270,7 +270,7 @@ class KMeans(KMeansClass, _CumlEstimator, _KMeansCumlParams):
                 concated = pd.concat(df_list)
             else:
                 # should be list of np.ndarrays here
-                concated = np.concatenate(df_list)
+                concated = np.array(np.concatenate(df_list), order='F')
 
             kmeans_object.fit(
                 concated,
@@ -283,13 +283,15 @@ class KMeans(KMeansClass, _CumlEstimator, _KMeansCumlParams):
                 f"iterations: {kmeans_object.n_iter_}, inertia: {kmeans_object.inertia_}"
             )
 
-            return {
+            res = {
                 "cluster_centers_": [
                     kmeans_object.cluster_centers_.to_numpy().tolist()
                 ],
                 "n_cols": params["n"],
-                "dtype": kmeans_object.dtype.name,
+                "dtype": str(kmeans_object.dtype.name),
             }
+            del kmeans_object
+            return res
 
         return _cuml_fit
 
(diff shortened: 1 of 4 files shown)


lijinf2:
Looks nice!

lijinf2:
Are there more commits on the way? I think the current shape looks good, assuming the run_tests.sh has got passed.

src/spark_rapids_ml/clustering.py
```diff
@@ -270,7 +270,13 @@ def _cuml_fit(
...
             else:
                 # should be list of np.ndarrays here
-                concated = np.concatenate(df_list)
+                rows = sum([arr.shape[0] for arr in df_list])
+                cols = df_list[0].shape[1]
+                d_type = df_list[0].dtype
+                concated_out = np.empty(shape=(rows, cols), order="F", dtype=d_type)
+                concated = np.concatenate(df_list, out=concated_out)
+                for df in df_list:
+                    del df
```
wbo4958, line 279: Could we put this into a utils function to be shared for all algos

wbo4958:
LGTM, a minor comment

src/spark_rapids_ml/clustering.py
eordentlich: Done.

src/spark_rapids_ml/utils.py
```diff
@@ -122,6 +122,26 @@ def build(cls, partition_rows: List[int], total_cols: int) -> "PartitionDescript
...
+def _concat_and_free(np_array_list: List[np.ndarray]) -> np.ndarray:
+    """
+    concatenates a list of compatible numpy arrays into a F ordered output array,
+    in a memory efficient way.
+    Note: frees list elements so do not reuse after calling.
+    """
+    rows = sum(arr.shape[0] for arr in np_array_list)
+    if len(np_array_list[0].shape) > 1:
+        cols = np_array_list[0].shape[1]
+        concat_shape: Tuple[int, ...] = (rows, cols)
```
leewyang, line 134: Not sure if this was intended to support more than two dimensions (because of the ellipsis), but `cols` as-written is only a single dimension, e.g. did you intend something more like this?
```
# 2+ dimensions
cols = np_array_list[0].shape[1:]
concat_shape: Tuple[int, ...] = (rows,) + cols
```

eordentlich: It declares concat_shape as a variable length tuple type due to the else branch assigning only a single component tuple.  mypy complained.

leewyang (dismissed):

leewyang (approved):

PR #148, opened 2023-03-14
fix docker ci

leewyang:
- specific version ranges in `requirements_dev.txt` for `black`, `isort`, and `mypy` to get more consistent behavior.
- fixed formatting errors for these latest formatter versions.
- modified `ci/lint_python.py` to run formatting (or type-checking) on all specified directories before failing.
- using same dataset paths for `--train_path` and `--transform_path` for now in `run_benchmark.sh` to fix "path does not exist" error.  (Pending PR from @wbo4958 should specify separate dataset paths).


eordentlich (approved):
👍

PR #150, opened 2023-03-15
fix mypy issues

leewyang:
Previously, we were checking the individual folders separately, but they really should be checked together.

### DIFF
--- a/src/spark_rapids_ml/clustering.py
+++ b/src/spark_rapids_ml/clustering.py
@@ -40,10 +40,8 @@ from spark_rapids_ml.core import (
     _CumlModel,
     _CumlModelSupervised,
 )
-from spark_rapids_ml.params import HasFeaturesCols, _CumlClass, _CumlParams
+from spark_rapids_ml.params import HasFeaturesCols, P, _CumlClass, _CumlParams
 from spark_rapids_ml.utils import get_logger
-
-P = TypeVar("P", bound=_CumlParams)
 
 
 class KMeansClass(_CumlClass):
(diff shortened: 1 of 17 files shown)


src/spark_rapids_ml/clustering.py
```diff
@@ -43,6 +43,8 @@
 from spark_rapids_ml.params import HasFeaturesCols, _CumlClass, _CumlParams
 from spark_rapids_ml.utils import get_logger
 
+P = TypeVar("P", bound=_CumlParams)
```
eordentlich, line 46: since this is repeated in several files, should it be in a common file?

leewyang: Yeah, thought about that, but it looks most people define locally per file.

leewyang: Hmm, looks like Spark actually uses a mix of local-defines and imports, so it's not consistent.
```
$ find python/pyspark -name "*.py" | xargs grep TypeVar | grep '"P"'
python/pyspark/ml/feature.py:P = TypeVar("P", bound=Params)
python/pyspark/ml/param/__init__.py:P = TypeVar("P", bound="Params")
$ find python/pyspark -name "*.py" | xargs grep "import P$"
python/pyspark/ml/base.py:from pyspark.ml.param import P
python/pyspark/ml/tree.py:    from pyspark.ml._typing import P
```

leewyang: Went ahead and updated with the `import P` method, since it's less repeated code.  Also updated `tree.py`, which wasn't failing any checks, just to be consistent.

eordentlich (approved):
👍

PR #158, opened 2023-03-16
register pytest.mark.compat marker

leewyang:
... to silence this warning in the test logs:
```
PytestUnknownMarkWarning: Unknown pytest.mark.compat - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
```


eordentlich (approved):
👍

PR #164, opened 2023-03-17
Databricks benchmark reproduce scripts

eordentlich:


benchmark/databricks/benchmark_utils.sh
leewyang, line 1: indentation is a mix of tabs and spaces in this file... should ideally be consistent.

benchmark/databricks/init-cpu.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# set portion of path below after /dbfs/ to dbfs zip file location
+SPARKCUML_ZIP=/dbfs/path/to/spark-rapids-ml.zip
+BENCHMARK_ZIP=/dbfs/path/to/benchmark.zip
+
+# install spark-cuml
+unzip ${SPARKCUML_ZIP} -d /databricks/python3/lib/python3.8/site-packages
```
leewyang, line 7: Should use `SPARK_RAPIDS_ML_ZIP` here (and above).

benchmark/databricks/init-pip-cuda-11.8.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# set portion of path below after /dbfs/ to dbfs zip file location
+SPARKCUML_ZIP=/dbfs/path/to/spark-rapids-ml.zip
```
leewyang, line 3: `SPARK_RAPIDS_ML_ZIP`

benchmark/databricks/init-pip-cuda-11.8.sh
```diff
@@ -0,0 +1,46 @@
...
+# install cudf and cuml
+#/databricks/python/bin/pip install cudf-cu11==${RAPIDS_VERSION} cuml-cu11==${RAPIDS_VERSION} --extra-index-url=https://pypi.ngc.nvidia.com
+/databricks/python/bin/pip install cudf-cu11==${RAPIDS_VERSION} \
+cuml-cu11==${RAPIDS_VERSION} \
+pylibraft-cu11==${RAPIDS_VERSION} \
+rmm-cu11==${RAPIDS_VERSION} \
+--extra-index-url=https://pypi.nvidia.com
+
+# install spark-cuml
+unzip ${SPARKCUML_ZIP} -d /databricks/python3/lib/python3.8/site-packages
```
leewyang, line 44: same

benchmark/databricks/run_benchmark.sh
```diff
@@ -0,0 +1,122 @@
...
+    rf_cpu_options=""
+fi
+num_runs=3
+
+sep="=================="
+
+echo
+echo "$sep algo: kmeans $sep"
+for i in `seq $kmeans_runs`; do run_bm kmeans \
+--num_runs 1 \
```
leewyang, line 48: Could just set `--num-runs $kmeans_runs` in theory (here and elsewhere in fine), unless you need to iterate in bash script explicitly.

leewyang (commented):
Some minor notes, otherwise LGTM

benchmark/databricks/run_benchmark.sh
eordentlich: Things become unstable for some reason with some of the algos, with task failures and longer run times.   We'll have to root cause that.   Could have added the inter-iteration-sleep in the python scripts but went this route.

benchmark/databricks/init-pip-cuda-11.8.sh
eordentlich: done all of the above

benchmark/databricks/benchmark_utils.sh
eordentlich: weird.  fixed.

leewyang (approved):

PR #165, opened 2023-03-20
Using namedtuple to replace global naming string

wbo4958:
We use the "global string" as the key of parameters with dictionary type passed from the core framework to the subclasses, which implies these "global strings" will be used in both core.py and other python files. What if we decide to change the naming for example "n" -> "num_cols", we need to change it everywhere. On the other hand, "global string" is easily mistaken. So I made this PR to use namedtuple to replace "global strings"

### DIFF
--- a/src/spark_rapids_ml/core.py
+++ b/src/spark_rapids_ml/core.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


wbo4958:
build

src/spark_rapids_ml/core.py
```diff
@@ -97,6 +94,12 @@
 Pred = namedtuple("Pred", ("prediction", "probability"))
 pred = Pred("prediction", "probability")
 
+# Global parameter alias used by core and subclasses.
+ParamAlias = namedtuple(
+    "ParamAlias", ("init", "handle", "num_cols", "part_sizes", "loop")
+)
+param_alias = ParamAlias("cuml_init", "handle", "n", "part_sizes", "loop")
```
leewyang, line 101: Did you intend to change "n" to "num_cols" here, since otherwise `param_alias.numl_cols` is still "n"?  Also, since we're making the change at this time, is there any reason why we shouldn't just set the values equal to the field names for now, e.g.
```
ParamAlias = namedtuple(
    "ParamAlias", ("cuml_init", "handle", "num_cols", "part_sizes", "loop")
)
param_alias = ParamAlias("cuml_init", "handle", "num_cols", "part_sizes", "loop")
```

lijinf2: Yeah, "n" has been used in the argument lists of cuml multi-gpu functions. "num_cols" reads more informative but replacement may trigger errors in calling cuml multi-gpu functions. 

It seems safe to replace "init" with "cuml_init".

leewyang (approved):

wbo4958:
build

PR #167, opened 2023-03-20
cache python dependencies in dev dockerfiles

leewyang:
Also:
- fix some cache directives.
- set min versions of key dependencies

Note: decided to leave `ci/Dockerfile` as-is to best reflect the build behavior w/o cached dependencies.  Also, the installation of pyspark  + other dependencies only took approx. 30s in a recent build log.


leewyang:
build

eordentlich (approved):
👍

leewyang:
build

PR #169, opened 2023-03-21
Use CICD-dedicated image

pxLi:
There is an internal pipeline to help re-build CICD images every 4 hours.

To avoid frequently updating image name, we change it to `ubuntu20.04-blossom-ci`,
the image content will be mostly based on the default ARGS and ENVs in dockerfile.


pxLi:
build

zhanga5 (approved):

PR #171, opened 2023-03-21
move python and jvm files into separate subdirectories

leewyang:


leewyang:
build

eordentlich:
Did you intend for the notebooks dir to be top level?

leewyang:
> Did you intend for the notebooks dir to be top level?

Yes, that was the intent (since most repos have an "examples" folder at the top level).  But we can move it if you want.

eordentlich:
Should the benchmark directory under python then be pulled out as well?  Maybe both go into a top level examples directory?   Also, do you know why the actual notebook files aren't showing as being moved?

leewyang:
Tried to pull the benchmark directory up, but that started raising issues w/ black/isort formatting.  As for the notebooks, not sure what you mean... they're still in their original locations, just updated the links in the READMEs.

eordentlich (approved):
👍

PR #177, opened 2023-03-23
another fix to array order

eordentlich:
- Missed that different algos have differing preferred orderings for the input arrays for fit (e.g. KMeans prefers C vs F) (warnings were suppressed in benchmarks in some cases - kmeans)

- Similar for transform.  KMeans prefers C while LinearRegression prefers F.   RandomForest on transform warns about either ordering, and is TBD to figure out.   Also, 1-d arrays as noted before are still problematic and are TBDs.

- Applied _concat_and_free to KNN with its preferred ordering.

- Added verbose option to benchmarks.   `eg. ./run_bench.sh kmeans --verbose`

### DIFF
--- a/python/benchmark/benchmark/base.py
+++ b/python/benchmark/benchmark/base.py
@@ -87,6 +87,12 @@ class BenchmarkBase:
             action="store_true",
             help="do not stop spark session when finished",
         )
+        self._parser.add_argument(
+            "--verbose",
+            type=int,
+            default=0,
+            help="verbose param for cuml classes",
+        )
 
         self._add_class_arguments()
         self._add_extra_arguments()
--- a/python/benchmark/benchmark/base.py
+++ b/python/benchmark/benchmark/base.py
@@ -29,6 +29,10 @@ from pyspark.sql.functions import col
 from .utils import WithSparkSession, to_bool, with_benchmark
 
 
+def _set_max_cuml_verbose() -> int:
+    return 7
+
+
 class BenchmarkBase:
     """Based class for benchmarking.
 
@@ -89,9 +93,10 @@ class BenchmarkBase:
         )
         self._parser.add_argument(
             "--verbose",
-            type=int,
+            action="store_const",
+            const=7,
             default=0,
-            help="verbose param for cuml classes",
+            help="set cuml logging to max verbose level",
         )
 
         self._add_class_arguments()
--- a/python/benchmark/benchmark/base.py
+++ b/python/benchmark/benchmark/base.py
@@ -29,10 +29,6 @@ from pyspark.sql.functions import col
 from .utils import WithSparkSession, to_bool, with_benchmark
 
 
-def _set_max_cuml_verbose() -> int:
-    return 7
-
-
 class BenchmarkBase:
     """Based class for benchmarking.
 
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -32,7 +32,6 @@ from spark_rapids_ml.tree import (
     _RandomForestEstimator,
     _RandomForestModel,
 )
-from spark_rapids_ml.utils import _ArrayOrder
 
 
 class _RFClassifierParams(_RandomForestClassifierParams, HasProbabilityCol):
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -32,6 +32,7 @@ from spark_rapids_ml.tree import (
     _RandomForestEstimator,
     _RandomForestModel,
 )
+from spark_rapids_ml.utils import _ArrayOrder
 
 
 class _RFClassifierParams(_RandomForestClassifierParams, HasProbabilityCol):
@@ -179,9 +180,8 @@ class RandomForestClassificationModel(
     ) -> Tuple[
         Callable[..., CumlT],
         Callable[[CumlT, Union[cudf.DataFrame, np.ndarray]], pd.DataFrame],
-        Literal["C", "F"],
     ]:
-        _construct_rf, _, array_order = super()._get_cuml_transform_func(dataset)
+        _construct_rf, _ = super()._get_cuml_transform_func(dataset)
 
         def _predict(rf: CumlT, pdf: Union[cudf.DataFrame, np.ndarray]) -> pd.Series:
             data = {}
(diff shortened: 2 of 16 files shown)


leewyang:
build

python/benchmark/benchmark/base.py
```diff
@@ -87,6 +87,12 @@ def __init__(self, argv: List[Any]) -> None:
             action="store_true",
             help="do not stop spark session when finished",
         )
+        self._parser.add_argument(
+            "--verbose",
+            type=int,
+            default=0,
+            help="verbose param for cuml classes",
```
lijinf2, line 94: Since the type is int instead of bool, need help information to know what different int values mean to the verbose argument.

python/src/spark_rapids_ml/core.py
```diff
@@ -614,7 +617,9 @@ def _cuml_transform(cuml_obj: CumlT, df: Union[pd.DataFrame, np.ndarray]) ->pd.D
                 ...
             ...
 
-            return _construct_cuml_object, _cuml_transform
+            array_order = "F"
```
lijinf2, line 620: array_order seems like a cuml param. Thinking it may be better to move it to the _CumlParams class.

eordentlich: It is not passed into the cuml object so not really consistent with that class as it stands.

python/benchmark/benchmark/base.py
eordentlich: Changed to boolean and set internal verbose level to max.

python/src/spark_rapids_ml/core.py
eordentlich: Created separate array_order methods for fit and transform (in case they are different) with most common case "F" as the default return value.

python/src/spark_rapids_ml/core.py
```diff
@@ -598,11 +599,13 @@ def _get_cuml_transform_func(
     ) -> Tuple[
         Callable[..., CumlT],
         Callable[[CumlT, Union[cudf.DataFrame, np.ndarray]], pd.DataFrame],
+        Literal["C", "F"],
```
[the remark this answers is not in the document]
eordentlich: Done.  (Though most got deleted due to implementation change).

python/benchmark/benchmark/base.py
```diff
@@ -29,6 +29,10 @@
 from .utils import WithSparkSession, to_bool, with_benchmark
 
 
+def _set_max_cuml_verbose() -> int:
+    return 7
+
+
```
leewyang, line 35: Is this used anywhere?

python/src/spark_rapids_ml/classification.py
```diff
@@ -32,6 +32,7 @@
     _RandomForestEstimator,
     _RandomForestModel,
 )
+from spark_rapids_ml.utils import _ArrayOrder
```
leewyang, line 35: imports seem no longer needed.

python/src/spark_rapids_ml/core.py
```diff
@@ -24,6 +24,7 @@
     Dict,
     Iterator,
     List,
+    Literal,
```
leewyang, line 27: no longer needed?

python/src/spark_rapids_ml/feature.py
```diff
@@ -15,7 +15,7 @@
 #
 
 import itertools
-from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, TypeVar, Union
```
leewyang, line 18: no longer needed?

python/src/spark_rapids_ml/tree.py
```diff
@@ -36,7 +36,7 @@
     param_alias,
 )
 from spark_rapids_ml.params import HasFeaturesCols, P, _CumlClass, _CumlParams
-from spark_rapids_ml.utils import _concat_and_free
+from spark_rapids_ml.utils import _ArrayOrder, _concat_and_free
```
leewyang, line 39: imports no longer needed?

python/tests/test_common_estimator.py
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
```
leewyang, line 17: no longer needed?

leewyang (approved):
minor cleanups, otherwise LGTM

lijinf2:
Look good!

python/src/spark_rapids_ml/core.py
```diff
@@ -471,6 +474,12 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
 
         return pipelined_rdd
 
+    def _fit_array_order(self) -> _ArrayOrder:
```
wbo4958, line 477: Can we merge _fit_array_order and _transform_array_order into _array_order and move into _CumlCommon?

python/src/spark_rapids_ml/feature.py
```diff
@@ -386,12 +386,15 @@ def _construct_pca() -> CumlT:
 
             pca.n_cols = self.n_cols
             pca.dtype = np.dtype(self.dtype)
+            # TBD: figure out why PCA warns regardless of array order here and for singular values
             pca.components_ = cudf_to_cuml_array(
```
wbo4958, line 390: Could we change it like 

``` python
order = self._fit_array_order()

np.array(self.components_, order=irder).astype(pca.dtype)
...

python/src/spark_rapids_ml/core.py
eordentlich: Then it would need to be overridden in both Model and Estimator which kind of defeats the purpose.     I don't see an algo specific common class that subclasses _CumlCommon .   I think will leave as is for now.

python/src/spark_rapids_ml/feature.py
eordentlich: This is actually a different case.  These are model params which have their own preferred order independent of the preferred order for the data.   I already set the coefficients order in pca to avoid the warning, but there seems to be a problem with 1d arrays hence the TBD.

eordentlich:
build

leewyang (approved):

PR #180, opened 2023-03-24
Retain feature columns when doing PCA transform

wbo4958:
This PR changed the previous PCA transform behavior that it outputs columnars for columnars input, to the single vector output.


leewyang (dismissed):

eordentlich (dismissed):
👍

wbo4958:
build

PR #181, opened 2023-03-24
update to cuml 23.02

leewyang:
Most of these changes tested fine w/o any issues, except:
- ~`Dockerfile.pip` had two test errors in `test_nearest_neighbors.py`.  Need to debug further.~ UPDATE: re-tried and it worked fine, so not sure why I saw errors before.
- Spot check of Databricks notebook had some intermittent barrier sync issues, which went away on retry.
- Did not run the Databricks benchmarks, since the changes are similar to the notebooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Note: Spark does not provide a k-NN implementation, but it does have an [LSH-bas
 
 ## Getting started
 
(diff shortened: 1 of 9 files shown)


leewyang:
build

eordentlich (approved):
👍

PR #184, opened 2023-03-28
Add toHost() to convert LinearRegressionModel to pyspark model.

wbo4958:


wbo4958:
build

wbo4958:
@leewyang @eordentlich Do we need to set PySpark LinearRegressionModel as the default model when doing predict/evaluate/transform?

python/src/spark_rapids_ml/core.py
```diff
@@ -586,6 +586,11 @@ def __init__(
         self.dtype = dtype
         self.n_cols = n_cols
 
+    def toHost(self) -> Model:
+        """Transform using pyspark ML model. For some reasons,
+        some algos can't be converted to pyspark ML Model"""
+        raise NotImplementedError()
```
leewyang, line 592: How about "Return the equivalent PySpark CPU model"?

And then, I'm wondering if we should rename `toHost()` to just `cpu()`, s.t. your example looks like:
```
model = LinearRegression().fit(xx)
# model.toHost().transform()
# model.toHost().predict()
model.cpu().transform()
model.cpu().predict()
```
... since `toHost()` makes me think of networking for some reason.  Not sure if anyone else has opinions on this?

wbo4958: Really good suggestion, will change it accordingly.

python/src/spark_rapids_ml/regression.py
```diff
@@ -443,18 +472,24 @@ def scale(self) -> float:
         """
         return 1.0
 
-    def evaluate(self, dataset: DataFrame) -> None:
-        """(Not supported) Evaluates the model on a test dataset."""
-        raise NotImplementedError
+    def predict(self, value: T) -> float:
+        """cuML doesn't support predicting 1 single sample.
+        Fall back to PySpark ML LinearRegressionModel"""
+        return self.cpu().predict(value)
```
eordentlich, line 478: This and the evaluate fall back are really nice benefits of this.

python/tests/test_linear_model.py
```diff
@@ -210,13 +218,36 @@ def assert_model(
         # train a model
         lr_model = lr.fit(df)
 
+        assert isinstance(lr_model.toHost(), SparkLinearRegressionModel)
```
eordentlich, line 221: A few more toHosts in this test.

python/src/spark_rapids_ml/core.py
eordentlich: Host/Device terminology is everywhere in cuda.    

cupy and cudf use library names (asNumpy and to_pandas) to move things back and forth.  Though cupy also has the .get() to move off device.  

as_spark_ml?  A little long.    cpu is good.

eordentlich (commented):
Nice.

wbo4958:
build

python/tests/test_linear_model.py
wbo4958: Thx. Done

python/src/spark_rapids_ml/regression.py
```diff
@@ -411,16 +424,32 @@ def __init__(
         dtype: str,
     ) -> None:
         super().__init__(dtype=dtype, n_cols=n_cols, coef_=coef_, intercept_=intercept_)
-        self.coef_ = coef_
-        self.intercept_ = intercept_
+        self._coef = coef_
+        self._intercept = intercept_
```
leewyang, line 428: Just curious why these are renamed?  Original intent was to just mirror the cuml attributes... and other classes follow the old naming scheme, so we should be consistent, if possible.

wbo4958: Sorry, I didn't realize it. I just would like to hide the internal variable. Let me change them back.

wbo4958:
build

eordentlich (approved):
👍

PR #187, opened 2023-03-30
Split deploy script [skip ci]

YanxuanLiu:
split `deploy.sh` script into `deploy.sh` and `docs.sh`

deploy script would be responsible to build whl package only. It will be called in blossom pipeline.

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -2,7 +2,7 @@
 # exit on first fail
 set -e
 
-# build pip package
+# build whl package
 pushd python
 pip install -r requirements_dev.txt && pip install -e .
 python -m build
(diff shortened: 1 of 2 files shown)


ci/deploy.sh
```diff
@@ -1,33 +1,9 @@
...
-
-# get version tag
-TAG=$(git describe --tag)
-if [[ $? != 0 ]]; then
-    echo "Can only deploy from a version tag."
-    exit 1
-fi
-
 # exit on first fail
 set -e
```
pxLi, line 3: nit: lets `set -ex` for easier debugging in logs

below the same

pxLi:
[skip ci] as will not be covered by pre-merge

pxLi (approved):

pxLi:
build

PR #190, opened 2023-03-31
[DO NOT MERGE] Refactor knn

leewyang:
Draft PR for discussion purposes.

KNN does not fit the traditional Spark ML Estimator + Model paradigm.  You don't really need to "train" it, and there really isn't any "model state" to be stored in the Model (other than `k`).  That said, we currently have an Estimator implementation that is "fit" on a DataFrame, which really just returns a Model with a reference to that DataFrame.  Then, the Model has APIs to compute nearest neighbors between that stored DataFrame to another DataFrame passed into the function arguments.  While this does work, it also leads to issues with saving the associated Estimator and Model instances, since the DataFrame reference is not serializable.

So, one option is to find a way to serialize a DataFrame reference cleanly.

This PR propose another option which removes the Estimator/Model base classes entirely, and just represents KNN as a simple class with stateless APIs.  This is most similar to the classes in [stats.py](https://github.com/apache/spark/blob/master/python/pyspark/ml/stat.py).

I also investigated basing on the Transformer class, but that API generally just "transforms" a single DataFrame at a time, so once again, we'd need to store a secondary DataFrame reference somewhere.

Anyhow, I've also staged the PR in multiple commits to simplify the review (with the first commit being the most important).  Please feel free to comment (and/or reject this entirely), since I'm not sure if we're losing something important with this change.

### DIFF
--- a/python/src/spark_rapids_ml/knn.py
+++ b/python/src/spark_rapids_ml/knn.py
@@ -13,12 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import asyncio
+from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
-import asyncio
 import numpy as np
 import pandas as pd
-from enum import IntEnum
 from pyspark.ml.functions import vector_to_array
 from pyspark.ml.linalg import VectorUDT
 from pyspark.ml.param.shared import (
@@ -40,19 +40,15 @@ from pyspark.sql.types import (
     StructType,
 )
 
-from spark_rapids_ml.core import (
-    CumlInputType,
-    _CumlCaller,
-    alias,
-    param_alias,
-)
+from spark_rapids_ml.core import CumlInputType, _CumlCaller, alias, param_alias
 from spark_rapids_ml.params import P, _CumlClass, _CumlParams
 from spark_rapids_ml.utils import _concat_and_free
 
 
-class Label(IntEnum):
-    DATA=0,
-    QUERY=1,
+class Label(Enum):
+    DATA = 0
+    QUERY = 1
+
 
 class NearestNeighborsClass(_CumlClass):
     @classmethod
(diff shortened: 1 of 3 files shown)


wbo4958:
Seems this PR has changed the signature of the `kneighbors`, so in that case, seems we don't need to store the `item_df: DataFrame` in KNearestNeighbors, which means we may directly return KNearestNeighborsModel in the `_fit` function of KNearestNeighbors. And this seems to conform to the spark estimator/model style. Please correct me.

lijinf2:
Yeah, the new APIs can avoid serializing a dataframe when saving/loading an estimator or model. The tradeoff is that fit(item_df) would be a no-op and may seem weird.  

Would it make sense to throw exceptions when save/load functions are called? Fit runs super fast (this is different from other algorithms), so it seems no need to store an estimator or model.

python/src/spark_rapids_ml/knn.py
```diff
@@ -226,64 +220,6 @@ class NearestNeighbors(
...
-        self._copy_cuml_params(model)  # type: ignore
-        return model
-
-    def _out_schema(self) -> Union[StructType, str]:  # type: ignore
-        """
-        This class overrides _fit and will not call _out_schema.
-        """
-        pass
-
-    def _get_cuml_fit_func(  # type: ignore
```
lijinf2, line 263: It is nice to get _out_shema and _get_cuml_fit_func removed.

python/src/spark_rapids_ml/knn.py
```diff
@@ -514,24 +455,10 @@ async def do_allGather() -> List[str]:
...
-    ]:
-        raise NotImplementedError(
-            "'_CumlModel._get_cuml_transform_func' method is not implemented. Use 'kneighbors' instead."
-        )
-
     def exactNearestNeighborsJoin(
         self,
         query_df: DataFrame,
+        item_df: DataFrame,
         distCol: str = "distCol",
```
lijinf2, line 462: One open question is which of query_df and item_df comes first in the argument list. 

Spark LSH has an API `approxNearestNeighbors(dataset, key, ...)` that puts item_df before the query vector.

leewyang:
> Would it make sense to throw exceptions when save/load functions are called? Fit runs super fast (this is different from other algorithms), so it seems no need to store an estimator or model.

Yeah, after all the comments (thanks, everyone!), and reviewing the sklearn and cuml APIs a bit more, I think this is the best solution.  Basically, leave as-is and just disable save/load.

PR #192, opened 2023-04-03
Auto infer to the prediction name for different algorithms

wbo4958:
Previously, we needed to manually add the prediction name for different algorithms when transforming. Eg, for PCA, we needed to specify "outputCol" while for others, we needed to specify "predictionCol". This is in-convenience, So I made this PR to auto-try the prediction name by matching HasPredictionCol and HasOutputCol.


wbo4958:
build

leewyang (approved):

PR #199, opened 2023-04-04
add docstring into NearestNeighbors class

lijinf2:
Also the example codes for vector and multi-col input types

### DIFF
--- a/python/src/spark_rapids_ml/knn.py
+++ b/python/src/spark_rapids_ml/knn.py
@@ -131,7 +131,7 @@ class _NearestNeighborsCumlParams(_CumlParams, HasInputCol, HasLabelCol, HasInpu
 
     def setIdCol(self: P, value: str) -> P:
         """
-        Sets the value of `id_col`. If not set, an id column will be added with column name `id`. The id column is used to specify nearest neighbor vectors by associated id value.
+        Sets the value of `id_col`. If not set, an id column will be added with column name `unique_id`. The id column is used to specify nearest neighbor vectors by associated id value.
         """
         self.set_params(id_col=value)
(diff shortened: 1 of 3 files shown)


lijinf2:
build

python/src/spark_rapids_ml/knn.py
```diff
@@ -221,6 +247,31 @@ class NearestNeighbors(
...
+    ...         (2, Vectors.dense([3.0, 3.0]),)]
+    >>> data_df = spark.createDataFrame(data, ["id", "features"])
+    >>> query = [(3, Vectors.dense([1.0, 1.0]),),
+    ...          (4, Vectors.dense([3.0, 3.0]),)]
+    >>> query_df = spark.createDataFrame(query, ["id", "features"])
+    >>> topk = 2
+    >>> gpu_knn = NearestNeighbors().setInputCol("features").setIdCol("id").setK(topk)
+    >>> gpu_model = gpu_knn.fit(data_df)
+
+    >>> from spark_rapids_ml.knn import NearestNeighbors
```
leewyang, line 264: I would remove this second import on the assumption that this is a single session.

leewyang: And maybe add comment lines explaining the two varieties, e.g.
```
# single column input
...
# multi-column input
```

lijinf2: did not delete the second import, thinking the example codes may be copied to a different session. 

array input is also single column so added the comments "vector column input" and "multi-column input".

lijinf2:
build

leewyang (approved):

PR #200, opened 2023-04-04
Support mixed model for RandomForestClassificationModel

wbo4958:
This PR implemented `cpu()` to convert Cuml RandomForestClassificationModel to Spark RandomForestClassificationModel, and a simple test to predict a single instance can pass. Still, this PR needs to be refactored.

For simplicity, This PR just returned the model JSON string (for CPU inference) alongside the model bytes (for GPU inference), but it increases the bandwidth when collecting from executor to driver which may cause perf issue. We need to find a way to combine them.

There's some difference between Cuml RandomForestClassificationModel and Spark RandomForestClassificationModel.

1. Cuml RandomForestClassificationModel doesn't include GiniStat in the InternalNode
2. Cuml RandomForestClassificationModel uses probability as the leaf value, while Spark RandomForestClassificationModel uses stats (how many instances each label contains), but we can convert the probs to the stats.
3. Cuml RandomForestClassificationModel doesn't have impurity information, but we can calculate it according to the leaf value.

Right now, this PR only supports Gini impurity, I will file a followup to support entropy and variance.

### DIFF
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -13,28 +13,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from typing import Any, Callable, Literal, Tuple, Type, Union
+import json
+from typing import Any, Callable, List, Optional, Tuple, Type, Union
 
 import cudf
 import numpy as np
 import pandas as pd
 from pyspark import Row
-from pyspark.ml.classification import _RandomForestClassifierParams
-from pyspark.ml.param.shared import HasProbabilityCol
+from pyspark.ml.classification import (
+    BinaryRandomForestClassificationSummary,
+    DecisionTreeClassificationModel,
+)
+from pyspark.ml.classification import (
+    RandomForestClassificationModel as SparkRandomForestClassificationModel,
+)
+from pyspark.ml.classification import (
+    RandomForestClassificationSummary,
+    _RandomForestClassifierParams,
+)
+from pyspark.ml.linalg import Vector
+from pyspark.ml.param.shared import HasProbabilityCol, HasRawPredictionCol
 from pyspark.sql import Column, DataFrame
 from pyspark.sql.functions import col
 from pyspark.sql.types import DoubleType, FloatType, IntegerType, IntegralType
 
-from spark_rapids_ml.core import CumlT, alias, pred
-from spark_rapids_ml.tree import (
+from .core import CumlT, alias, pred
+from .tree import (
     _RandomForestClass,
     _RandomForestCumlParams,
     _RandomForestEstimator,
     _RandomForestModel,
 )
+from .utils import _get_spark_session, java_uid, translate_trees
 
 
-class _RFClassifierParams(_RandomForestClassifierParams, HasProbabilityCol):
+class _RFClassifierParams(
+    _RandomForestClassifierParams, HasProbabilityCol, HasRawPredictionCol
+):
     def __init__(self, *args: Any):
         super().__init__(*args)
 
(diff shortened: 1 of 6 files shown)


leewyang:
> For simplicity, This PR just returned the model JSON string (for CPU inference) alongside the model bytes (for GPU inference), but it increases the bandwidth when collecting from executor to driver which may cause perf issue. We need to find a way to combine them.

Wonder if you can avoid serializing the model JSON string by just using [treelite.dump_as_json()](https://treelite.readthedocs.io/en/latest/treelite-api.html#treelite.Model.dump_as_json) on the driver side, after collecting the trees from the executors.  Not sure if that JSON format is compatible/sufficient, but it would avoid serializing the model essentially twice (in treelite and JSON formats).

python/tests/test_random_forest.py
```diff
@@ -610,3 +610,31 @@ def test_random_forest_regressor_spark_compat(
...
+        rf = RandomForestClassifier(
+            num_workers=1, numTrees=1, maxDepth=2, labelCol="indexed", seed=42
+        )
+        model: RandomForestClassificationModel = rf.fit(td)
+        z = model.cpu()
+
+        from pyspark.ml.common import _py2java
+
+        example = _py2java(spark.sparkContext, Vectors.dense(1.0))
+        print(z.predict(example))
```
lijinf2, line 640: Need assert statements to ensure the predicted results are the same as pyspark.

wbo4958:
build

wbo4958:
> > For simplicity, This PR just returned the model JSON string (for CPU inference) alongside the model bytes (for GPU inference), but it increases the bandwidth when collecting from executor to driver which may cause perf issue. We need to find a way to combine them.
> 
> Wonder if you can avoid serializing the model JSON string by just using [treelite.dump_as_json()](https://treelite.readthedocs.io/en/latest/treelite-api.html#treelite.Model.dump_as_json) on the driver side, after collecting the trees from the executors. Not sure if that JSON format is compatible/sufficient, but it would avoid serializing the model essentially twice (in treelite and JSON formats).

Thx @leewyang, it seems that cuML model is not compatible with treelite model.

``` console
treelite_json = cu_rf.convert_to_treelite_model().dump_as_json()
AttributeError: 'cuml.fil.fil.TreeliteModel' object has no attribute 'dump_as_json'
```

python/tests/test_random_forest.py
wbo4958: Hmm. The predictions between cuML and pyspark may be different. It's hard to compare the prediction results.

leewyang:
Wonder if this might be useful: https://github.com/rapidsai/cuml/issues/3853
Sounds like it'd require an extra serialize/deserialize to disk to get it to the actual treelite model class, but presumably this would only need to be done on the driver.

python/src/spark_rapids_ml/classification.py
```diff
@@ -210,3 +284,68 @@ def hasSummary(self) -> bool:
...
+        """
+        return self.cpu().predictRaw(value)
+
+    def predictProbability(self, value: Vector) -> Vector:
+        """
+        Predict the probability of each class given the features.
+        """
+        return self.cpu().predictProbability(value)
+
+    def evaluate(
```
eordentlich, line 312: We should consider a gpu version of this where we have all the supporting metrics needed to construct the summary also implemented on gpu.

eordentlich: In future work.

wbo4958:
> Wonder if this might be useful: [rapidsai/cuml#3853](https://github.com/rapidsai/cuml/issues/3853) Sounds like it'd require an extra serialize/deserialize to disk to get it to the actual treelite model class, but presumably this would only need to be done on the driver.
> 
> Otherwise, LGTM. (Also, pls remove "[DRAFT]" if ready for review).

Really good, Seems we can just return the json strings of treelite in the executor side according to the example of that PR.  Great. Thx, I will file a follow up for it.

python/src/spark_rapids_ml/classification.py
wbo4958: Good suggestion, Let me file a task for it.

leewyang (approved):

wbo4958:
build

wbo4958:
build

PR #216, opened 2023-04-10
instructions for running notebooks on Dataproc

leewyang:
Also updated `spark-compat.ipynb` to remove outputs and simplify save/load operations (to make it easier to work with in Dataproc).

### DIFF
--- a/notebooks/dataproc/spark_rapids_ml.sh
+++ b/notebooks/dataproc/spark_rapids_ml.sh
@@ -15,5 +15,8 @@ pylibraft-cu11==${RAPIDS_VERSION} \
 rmm-cu11==${RAPIDS_VERSION} \
 --extra-index-url=https://pypi.nvidia.com
 
+# rapids pip package patch: link ucx libraries in ucx-py to default location searched by raft_dask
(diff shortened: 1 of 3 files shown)


leewyang:
skipping build, since these files won't affect it at all.

notebooks/dataproc/spark_rapids_ml.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# patch existing
+pip install --ignore-installed "llvmlite<0.40,>=0.39.0dev0" "numba>=0.56.2"
+
+# install cudf and cuml
+pip install cudf-cu11==${RAPIDS_VERSION} \
+cuml-cu11==${RAPIDS_VERSION} \
+pylibraft-cu11==${RAPIDS_VERSION} \
+rmm-cu11==${RAPIDS_VERSION} \
+--extra-index-url=https://pypi.nvidia.com
```
eordentlich, line 16: Heads up that for ucx in knn you will most likely still need to patch the pip install with the symlinks as in the Databricks init script.   If that doesn't work due to permissions, there are other ways to link within the site-packages.

eordentlich (approved):
👍

PR #222, opened 2023-04-12
RFclassifier: Support entropy impurity when convering cuml trees to Spark

wbo4958:
It depends on #200 which is already merged. Now it's open to review.

This PR supports entropy impurity for RandomForestClassificationModel when converting cuml trees to spark. And it adds a checking that compares if the prediction results are the same using Spark-Rapids-ML RandomForestClassificationModel and the CPU RandomForestClassificationModel model converted by `cpu()` API.


leewyang:
:+1:

wbo4958:
build

leewyang (approved):

PR #228, opened 2023-04-14
[hot fix] set pyspark upper bond dependency [skip ci]

wbo4958:
Recently, Spark has released 3.4.0 which has blocked our testing. After fixing the issues, we will update the pyspark upper bond dependency.


pxLi (approved):

pxLi:
build

PR #230, opened 2023-04-14
Support mixed model for Rf regression model

wbo4958:
This is the final PR to support random forest mixed model. Since the model size passed from executor to driver resulted in OOM of java space, I just increased the driver/executor memory to bypass this issue. Later I will check how to improve this part.


wbo4958:
build

leewyang (approved):

PR #234, opened 2023-04-17
fix mapping of featureSubsetStrategy

eordentlich:


python/src/spark_rapids_ml/params.py
```diff
@@ -407,16 +409,8 @@ def _set_cuml_value(self, k: str, v: Any) -> None:
             self._cuml_params[k] = v
         else:
             # value map exists
-            supported_values = set([x for x in value_map[k].values() if x is not None])
-            if v in supported_values:
-                # already a valid value
-                self._cuml_params[k] = v
+            mapped_v = value_map[k](v)
```
wbo4958, line 412: If touching this common part, Seems, I guess, we need to update all the mappings in the subclass?

python/src/spark_rapids_ml/regression.py
```diff
@@ -89,10 +89,21 @@ def _param_mapping(cls) -> Dict[str, Optional[str]]:
...
+            "loss": lambda x: {
+                "squaredError": "squared_loss",
+                "huber": None,
+                "squared_loss": "squared_loss",
+            }.get(x, None),
+            "solver": lambda x: {
+                "auto": "eig",
+                "normal": "eig",
+                "l-bfgs": None,
+                "eig": "eig",
```
wbo4958, line 105: Do we need to map "eig" to "eig"?

python/src/spark_rapids_ml/regression.py
```diff
@@ -89,10 +89,21 @@ def _param_mapping(cls) -> Dict[str, Optional[str]]:
...
+    def _param_value_mapping(
+        cls,
+    ) -> Dict[str, Callable[[str], Union[None, str, float, int]]]:
         return {
-            "loss": {"squaredError": "squared_loss", "huber": None},
-            "solver": {"auto": "eig", "normal": "eig", "l-bfgs": None},
+            "loss": lambda x: {
+                "squaredError": "squared_loss",
+                "huber": None,
+                "squared_loss": "squared_loss",
```
wbo4958, line 99: Do we need to map "squared_loss" to "squared_loss"

leewyang: Think this is now required since the original check against `supported_values` has been replaced with the function, which needs to map every possible input to an output now.

leewyang:
build

eordentlich:
build

python/src/spark_rapids_ml/regression.py
eordentlich: That's correct.  The previous logic with dicts allowed specifying the mapped output values as input in which case they would be passed through, but with the more general function based logic, it is no longer possible to determine all possible outputs of the function.

python/src/spark_rapids_ml/params.py
eordentlich: I attempted to do this.   Did I miss something?

eordentlich:
build

eordentlich:
build

python/src/spark_rapids_ml/params.py
leewyang: Sorry, I interpreted this as, "all of the other subclasses need to be updated"... but upon further review, it looks like you've already updated all occurrences of `_param_value_mapping` (which does not exist for all subclasses).  So unless @wbo4958 meant something else, I think this is OK.

python/src/spark_rapids_ml/regression.py
leewyang: We should probably update the pydocs for `_param_value_mapping` in core.py to clarify this (requiring all valid inputs to be handled/mapped).

wbo4958 (dismissed):
LGTM

python/src/spark_rapids_ml/regression.py
eordentlich: Check just updated pydocs.

eordentlich:
build

leewyang (approved):

PR #240, opened 2023-04-19
revert db init script change

eordentlich:
It seemed pinning these dependencies wasn't necessary so removed them, but I just tried the benchmarks again and 23.04 versions of `rmm` and `pylibraft` were being pulled in causing the benchmarks to fail.

We still need 22.12 rapids version overall till we figure a10 issue on db so leaving that.


eordentlich:
No need to build for these changes.

leewyang:
Do we need to change for Dataproc init scripts as well?

leewyang (approved):

lijinf2:
build

PR #242, opened 2023-04-20
Support deployment to AWS EMR

lijinf2:
Got all notebooks (spark-compat, pca, linear-regression, knn, and kmeans) tested and passed in AWS EMR


notebooks/aws-emr/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+  ```
+  aws ec2 describe-subnets
+  export SUBNET_ID=<your_SubnetId>
+  ```
+
+- Create a cluster with at least two single-gpu TASK workers. You will obtain a ClusterId in terminal.  
+
+  ```
+
```
leewyang, line 39: Remove blank line.

notebooks/aws-emr/README.md
```diff
@@ -0,0 +1,74 @@
...
+  --applications Name=Hadoop Name=Livy Name=Spark Name=JupyterEnterpriseGateway \
+  --service-role EMR_DefaultRole \
+  --ec2-attributes SubnetId=${SUBNET_ID},InstanceProfile=EMR_EC2_DefaultRole \
+  --instance-groups InstanceGroupType=MASTER,InstanceCount=1,InstanceType=m4.4xlarge \
+                    InstanceGroupType=CORE,InstanceCount=1,InstanceType=g4dn.2xlarge \
+                    InstanceGroupType=TASK,InstanceCount=1,InstanceType=g4dn.2xlarge \
+                    InstanceGroupType=TASK,InstanceCount=1,InstanceType=g4dn.2xlarge \
+  --configurations file://${CUR_DIR}/init-configurations.json \
+  --bootstrap-actions Name='Spark Rapids ML Bootstrap action',Path=s3://${S3_BUCKET}/init-bootstrap-action.sh
+  ```
```
leewyang, line 55: Might be nice to explain how MASTER, CORE, TASK map to the various spark nodes (in case folks want to adjust these in the future).  For instance, I'm wondering why we need three GPU nodes if the driver wasn't using GPU?

leewyang (commented):
Minor comments, otherwise LGTM

notebooks/aws-emr/README.md
```diff
@@ -0,0 +1,74 @@
...
+  aws configure
+  AWS Access Key ID [None]: <your_access_key>
+  AWS Secret Access Key [None]: <your_secret_access_key>
+  Default region name [None]: <region-code>
+  Default output format [None]: <json>
+  ```
+- Create an S3 bucket if you don't already have one.
+  ```
+  export S3_BUCKET=<your_gcs_bucket_name>
```
eordentlich, line 15: gcs -> s3

notebooks/aws-emr/README.md
```diff
@@ -0,0 +1,74 @@
...
+  export CUR_DIR=$(pwd)
+
+  aws emr create-cluster \
+  --name ${CLUSTER_NAME} \
+  --release-label emr-6.10.0 \
+  --applications Name=Hadoop Name=Livy Name=Spark Name=JupyterEnterpriseGateway \
+  --service-role EMR_DefaultRole \
+  --ec2-attributes SubnetId=${SUBNET_ID},InstanceProfile=EMR_EC2_DefaultRole \
+  --instance-groups InstanceGroupType=MASTER,InstanceCount=1,InstanceType=m4.4xlarge \
+                    InstanceGroupType=CORE,InstanceCount=1,InstanceType=g4dn.2xlarge \
```
eordentlich, line 50: This could be cpu node.   m4.2xlarge.

notebooks/aws-emr/README.md
```diff
@@ -0,0 +1,74 @@
...
+                    InstanceGroupType=TASK,InstanceCount=1,InstanceType=g4dn.2xlarge \
+                    InstanceGroupType=TASK,InstanceCount=1,InstanceType=g4dn.2xlarge \
```
eordentlich, line 52: Can we have one of these with InstanceCount=2?

notebooks/aws-emr/init-configurations.json
```diff
@@ -0,0 +1,87 @@
...
+		"spark.sql.execution.sortBeforeRepartition":"false",
+		"spark.sql.execution.arrow.pyspark.enabled":"true",
+		"spark.sql.execution.arrow.maxRecordsPerBatch":"100000",
+		"spark.sql.cache.serializer com.nvidia.spark":"ParquetCachedBatchSerializer",
+		"spark.pyspark.python":"python3.9",
+		"spark.pyspark.driver.python":"python3.9",
+		"spark.dynamicAllocation.enabled":"false",
+		"spark.driver.memory":"20g",
+		"spark.rpc.message.maxSize":"512",
+		"spark.executorEnv.CUPY_CACHE_DIR":"/tmp/.cupy",
```
eordentlich, line 76: Interesting.  Do we have this elsewhere?   What happens if it is not set?

notebooks/kmeans.ipynb
```diff
@@ -52,12 +52,12 @@
...
     "dtype='float32'\n",
-    "from cuml.datasets import make_blobs\n",
+    "from sklearn.datasets import make_blobs\n",
     "data, _ = make_blobs(\n",
-    "        n_rows, n_cols, n_clusters_data, cluster_std=cluster_std, random_state=0, dtype=dtype\n",
+    "        n_samples=n_rows, n_features=n_cols, centers=n_clusters_data, cluster_std=cluster_std, random_state=0\n",
     "    )  # make_blobs creates a random dataset of isotropic gaussian blobs.\n",
     "\n",
-    "data = data.get()"
+    "data = data.astype(np.float32)"
```
eordentlich, line 60: np.float32 -> dtype ?

notebooks/aws-emr/README.md
lijinf2: Good suggestion. Just added a few words to explain why three GPU instances to get two workers.

notebooks/aws-emr/init-configurations.json
lijinf2: Yeah, this also appears in Dataproce. 

It not set, there will be a permission error writing to /home/.cupy.

notebooks/aws-emr/README.md
leewyang: So EMR is _randomly_ picking either a CORE or a TASK node to run JupyterLab?  I would assume that TASK nodes would be reserved for the executors, so JupyerLab would run on CORE (which could theoretically be a CPU machine).

lijinf2: Yeah, printed out the node can be CORE or TASK. There could be a way to specifically assign the node.

notebooks/aws-emr/init-configurations.json
eordentlich: We'll have to look into whether isolation is needed for this directory in the case of concurrent spark jobs in the same cluster.

notebooks/aws-emr/README.md
eordentlich: CORE nodes run hdfs while TASK nodes do not.   That seems to be the main distinction and it doesn't constrain where the driver goes.   I guess we can force the driver to land on a cpu only node by giving that node more memory than the other nodes and setting the driver memory such that yarn must schedule it to that node.   The executors can then land on the gpu nodes based on the gpu resource requirements specified for executors.

eordentlich: Ideally yarn should be 'smart' and not schedule a driver that doesn't request gpu resource to a gpu node when adequate resources are available on a cpu node.

notebooks/aws-emr/init-configurations.json
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	]
+},
+{
+	"Classification":"spark-defaults",
+	"Properties":{
+		"spark.plugins":"com.nvidia.spark.SQLPlugin",
+		"spark.sql.sources.useV1SourceList":"",
+		"spark.executor.resource.gpu.discoveryScript":"/usr/lib/spark/scripts/gpu/getGpusResources.sh",
```
eordentlich, line 47: Can you check formatting on this file?  (tabs vs spaces, etc)?  Looks like entries are misaligned.

eordentlich:
@lijinf2 Can you address remaining few comments?   **No need to pass build** once it gets an approval to merge since no changes go through ci/cd testing.

notebooks/aws-emr/README.md
lijinf2: (1) The cost for large memory can be higher than small memory + GPU. AWS pricing shows m4.4xlarge (64G) costs 0.8/h, and g4dn.2xlarge (32G + GPU) costs  0.75/h. 

(2) Yeah, not sure if this favorable feature of selecting CPU/GPU nodes is supported. 

(3) Tried other option to avoid GPU driver like node-label https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-master-core-task-nodes.html. Did not get it to work.

eordentlich (approved):
Nice work!

notebooks/aws-emr/init-configurations.json
lijinf2: Good question. The CUPY_CACHE_DIR comes from https://docs.cupy.dev/en/stable/reference/environment.html. 
Not sure if CUPY official handles concurrent cupy tasks. Need to try and find out. Created a ticket.

notebooks/aws-emr/init-configurations.json
lijinf2: Replaced all tab with four space using auto alignment vim command 'gg=G'.

PR #243, opened 2023-04-20
Run spark-rapids-ml on spark 3.4

wbo4958:
This PR makes spark-rapids-ml run against spark 3.4 by resolving the format checking. And it updates the requirements to the latest spark 3.4.0


wbo4958:
build

leewyang (approved):

eordentlich:
This is great.  With this, we can now test on Databricks 13.0 runtime which also has a better cuda setup and we may not need to install the cudatoolkit (which takes a while).

PR #245, opened 2023-04-21
update rapids dependencies to 23.4

eordentlich:
left spark rapids jar version at 23.02 as it is not yet updated to 23.04 in maven central

ran tests and benchmark ci docker image and on Databricks


eordentlich:
build

leewyang (approved):

PR #248, opened 2023-04-24
updating remaining 23.02 in dataproc README; fix long knn test

eordentlich:
Updating knn test to compare cuml single node and spark rapids ml to each other, instead of to 0 for self-distance check.   Seems in 23.04 single node and mnmg knn now use similar code paths that use less accurate but faster distance computation:  https://docs.rapids.ai/api/cuml/stable/api/#cuml.neighbors.NearestNeighbors.kneighbors (see `two_pass_precision`) option.


eordentlich:
build

lijinf2 (approved):
Nice.

PR #250, opened 2023-04-26
update spark rapids versions to 23.04

eordentlich:
Now that spark-rapids 23.04 jar is published to maven we can update our scripts.

Tested on DB benchmarks.

No need for CI build for this PR.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -98,7 +98,7 @@ ENV RAFT_PATH=/raft
 
 ### END OF CACHE ###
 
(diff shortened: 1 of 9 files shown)


leewyang (dismissed):

leewyang:
Might want to update [this](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.04/notebooks/aws-emr/init-bootstrap-action.sh#L11) as well... but not sure if still in progress.

leewyang (approved):

PR #256, opened 2023-04-28
improve api docs

eordentlich:
- adapted sphinx set up from Pyspark for more similar look to api docs.
- sphinx logic to skip unsupported attributes and methods.
- avoid documenting non-user facing classes and apis

TBD - Additional work is required to:
- override Param attributes for which not all values are supported, e.g. solver for LinearRegression
- class constructor inputs should appear in documentation
- no unsupported methods (besides getters and setters for unsupported attributes which are handled in this PR) should appear in docs
- delete unsupported methods and attributes at runtime in code, or override with not implemented exceptions.
- document cuml versions of parameters 
- add possibly missing setters


eordentlich:
build

python/src/spark_rapids_ml/utils.py
```diff
@@ -28,6 +28,27 @@
...
 
 
+def _method_names_from_param(spark_param_name: str) -> List[str]:
+    cap = spark_param_name[0].upper() + spark_param_name[1:]
+    getter = f"get{cap}"
+    setter = f"set{cap}"
+    return [getter, setter]
+
+
+def _unsupported_methods_attributes(clazz: Any) -> Set[str]:
```
leewyang, line 38: Probably should add pydoc for this function (and maybe also for above).

docs/source/spark_rapids_ml.rst
```diff
@@ -1,93 +1,88 @@
...
+..  Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+
+Spark Rapids ML
+=======================
```
leewyang, line 20: Not sure if line length needs to match (but minimally, seems like convention)

leewyang (commented):
Still need to review output of doc build.

leewyang:
Is there any way we can remove the extra level (spark_rapids_ml -> Spark Rapids ML) in the main page?
Screenshot 2023-05-01 at 6 00 10 PM

leewyang:
Upon spot check, it looks like the PCA docs now mostly match what you'd see in Jupyter auto-complete, which is good.  

However, the PCAModel docs looks like it might be missing some of the "cuml" attributes, e.g. `components_` and `explained_variance_ratio_`, which are actually wrapped by "spark" attributes like `pc` and `explainedVariance`.  That said, I think this is fine as-is.

leewyang (dismissed):

eordentlich:
Addressed comments.

Updating PR description for TBDs.

leewyang (approved):

eordentlich:
build

PR #258, opened 2023-05-01
fix typo in local notebook ssh tunnel command and add a clarification

eordentlich:


eordentlich:
no build needed.

notebooks/README.md
```diff
@@ -28,9 +28,9 @@ To run notebooks using Spark local mode on a server with one or more NVIDIA GPUs
     ```bash
     export USER=<your_username>
     export HOST=<your_hostname>
-    ssh -A -L 8888:127.0.0.1:8888 -L 4040:127.0.0.1:4040 -L $USER@$HOST
+    ssh -A -L 8888:127.0.0.1:8888 -L 4040:127.0.0.1:4040 $USER@$HOST
     ```
-    Then, browse to the `127.0.0.1` URL printed by the command in step 4.
+    Then, browse to the `127.0.0.1` URL printed by the command in step 4.   Note that a tunnel is also opened to the Spark UI server on port 4040.  Once a notebook is opened, you can view it by browsing to `http://127.0.0.1:4040` in another tab or window.
```
leewyang, line 33: Minor nit: if you don't backquote the http URL, I _think_ it would be a clickable link, but not really necessary.

leewyang (dismissed):

notebooks/README.md
eordentlich: Nice.   

While at it changed the host and user env names to avoid confusion with local versions.

leewyang (approved):

PR #263, opened 2023-05-11
Fix a bug cuml parameters have not been updated

wbo4958:
When users call estimator.fit(df, {parameters}), it first copies the parameters into estimator, but it can't propagated them into cuml parameters.

### DIFF
--- a/python/src/spark_rapids_ml/params.py
+++ b/python/src/spark_rapids_ml/params.py
@@ -178,17 +178,23 @@ class _CumlParams(_CumlClass, Params):
 
     def copy(self: P, extra: Optional["ParamMap"] = None) -> P:
         # override this function to update cuml_params if possible
+        instance: P = super().copy(extra)
+        cuml_params = instance.cuml_params.copy()
+
         if isinstance(extra, dict):
             for param, value in extra.items():
                 if isinstance(param, Param):
-                    self._set_cuml_param(param.name, value, silent=False)
+                    name = instance._get_cuml_param(param.name, silent=False)
+                    if name is not None:
+                        cuml_params[name] = value
                 else:
                     raise TypeError(
                         "Expecting a valid instance of Param, but received: {}".format(
                             param
                         )
                     )
-        return super().copy(extra)
+        instance._cuml_params = cuml_params
+        return instance
 
     def initialize_cuml_params(self) -> None:
         """
--- a/python/src/spark_rapids_ml/params.py
+++ b/python/src/spark_rapids_ml/params.py
@@ -15,12 +15,25 @@
 #
 
 from abc import abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 from pyspark.ml.param import Param, Params, TypeConverters
 from pyspark.sql import SparkSession
 
 from .utils import _is_local
+
+if TYPE_CHECKING:
+    from pyspark.ml._typing import ParamMap
 
 P = TypeVar("P", bound="_CumlParams")
 
@@ -162,6 +175,20 @@ class _CumlParams(_CumlClass, Params):
     @num_workers.setter
     def num_workers(self, value: int) -> None:
         self._num_workers = value
+
+    def copy(self: P, extra: Optional["ParamMap"] = None) -> P:
+        # override this function to update cuml_params if possible
+        if isinstance(extra, dict):
+            for param, value in extra.items():
+                if isinstance(param, Param):
+                    self._set_cuml_param(param.name, value, silent=False)
+                else:
+                    raise TypeError(
+                        "Expecting a valid instance of Param, but received: {}".format(
+                            param
+                        )
+                    )
+        return super().copy(extra)
 
     def initialize_cuml_params(self) -> None:
         """
@@ -338,6 +365,27 @@ class _CumlParams(_CumlClass, Params):
 
         return num_workers
 
+    def _get_cuml_param(self, spark_param: str, silent: bool = True) -> Optional[str]:
+        param_map = self._param_mapping()
+
+        if spark_param in param_map:
+            cuml_param = param_map[spark_param]
+            if cuml_param is None:
+                if not silent:
+                    # if Spark Param is mapped to None, raise error
+                    raise ValueError(
+                        f"Spark Param '{spark_param}' is not supported by cuML."
+                    )
+            elif cuml_param == "":
+                # if Spark Param is mapped to empty string, warn and continue
+                if not silent:
+                    print(f"WARNING: Spark Param '{spark_param}' is not used by cuML.")
+                cuml_param = None
+
+            return cuml_param
+        else:
+            return None
+
     def _set_cuml_param(
         self, spark_param: str, spark_value: Any, silent: bool = True
     ) -> None:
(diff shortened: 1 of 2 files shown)


wbo4958:
build

python/src/spark_rapids_ml/params.py
```diff
@@ -163,6 +176,20 @@ def num_workers(self) -> int:
     def num_workers(self, value: int) -> None:
         self._num_workers = value
 
+    def copy(self: P, extra: Optional["ParamMap"] = None) -> P:
+        # override this function to update cuml_params if possible
+        if isinstance(extra, dict):
+            for param, value in extra.items():
+                if isinstance(param, Param):
+                    self._set_cuml_param(param.name, value, silent=False)
```
eordentlich, line 184: Should the extra items update the original or the copy?   Looks like it is updating the original here.

python/src/spark_rapids_ml/params.py
```diff
@@ -338,6 +365,27 @@ def _infer_num_workers(self) -> int:
 
         return num_workers
 
+    def _get_cuml_param(self, spark_param: str, silent: bool = True) -> Optional[str]:
```
eordentlich, line 368: Why factor this out?  It was only called in one place.

python/tests/test_common_estimator.py
```diff
@@ -425,6 +425,15 @@ def assert_model(model: SparkRapidsMLDummyModel) -> None:
...
         assert_model(model_loaded)
 
+        dummy2 = dummy.copy()
+        assert dummy2.cuml_params["a"] == 100
+        with pytest.raises(
+            Exception,
+            match="assert {'a': 9876, 'k': 4, 'x': 40.0} == {'a': 100, 'k': 4, 'x': 40.0}",
+        ):
+            dummy2.fit(df, {dummy2.alpha: 9876})
+        assert dummy2.cuml_params["a"] == 9876
```
eordentlich, line 435: Not clear this is desired behavior, that original is updated with overriding param values.

eordentlich (commented):
Should we just apply the spark params to cuml params mapping on the fly just before invoking fit and transform instead of maintaining a separate cuml_params dictionary?

The dictionary would only be used for cuml params set directly in the constructor and would be overridden by any corresponding Spark Param.

python/tests/test_common_estimator.py
leewyang: I agree that this doesn't look right.  I'd expect something like the following:
```
assert dummy2.cuml_params["a"] == 100
model2 = dummy2.fit(df, {dummy2.alpha: 9876})
assert dummy2.cuml_params["a"] == 100
assert model2.cuml_params["a"] == 9876
```

Basically, the extra param map should produce a temporary copy of the estimator which is then fit on the dataframe to produce `model2`.  This is how the Spark implementation behaves.

python/src/spark_rapids_ml/params.py
leewyang: Think this needs to update a copy.

python/src/spark_rapids_ml/params.py
leewyang: I can _maybe_ see a case for symmetry of get/set, but I'm not sure we'd use `_get_cuml_param` anywhere else.

python/tests/test_common_estimator.py
leewyang: Also note that in this line, `dummy2.alpha` doesn't set the value, it's just a Param (descriptor) paired with a value, so it could easily also be `dummy.alpha: 9876`.  Either way, it wouldn't update `dummy` or `dummy2`, but the hidden/temporary copy of `dummy2`.
```
model2 = dummy2.fit(df, {dummy2.alpha: 9876})
```

python/src/spark_rapids_ml/params.py
wbo4958: Actually, itt's needed in my followup PR. :)

python/tests/test_common_estimator.py
wbo4958: Yes, you're right. I changed the spark behaviors. I will fix that.

python/src/spark_rapids_ml/params.py
wbo4958: yes, it's a bug, I just fixed it.

wbo4958:
> Should we just apply the spark params to cuml params mapping on the fly just before invoking fit and transform instead of maintaining a separate cuml_params dictionary?
> 
> The dictionary would only be used for cuml params set directly in the constructor and would be overridden by any corresponding Spark Param.


Hmm, yes, the idea of updating on the fly is ok for the training/transform. But the PR overriding the copy can also fix the copy() API itself issue. Please take a look at the new test I added.

wbo4958:
build

python/src/spark_rapids_ml/params.py
```diff
@@ -163,6 +176,26 @@ def num_workers(self) -> int:
     def num_workers(self, value: int) -> None:
         self._num_workers = value
 
+    def copy(self: P, extra: Optional["ParamMap"] = None) -> P:
+        # override this function to update cuml_params if possible
+        instance: P = super().copy(extra)
```
leewyang, line 181: Not sure we want `super().copy()` here, since Spark generally uses `self.copy()`.

python/tests/test_common_estimator.py
```diff
@@ -429,10 +437,11 @@ def assert_model(model: SparkRapidsMLDummyModel) -> None:
         assert dummy2.cuml_params["a"] == 100
         with pytest.raises(
             Exception,
-            match="assert {'a': 9876, 'k': 4, 'x': 40.0} == {'a': 100, 'k': 4, 'x': 40.0}",
+            match="assert {'a': 9876.0, 'k': 4, 'x': 40.0} == {'a': 100, 'k': 4, 'x': 40.0}",
         ):
-            dummy2.fit(df, {dummy2.alpha: 9876})
-        assert dummy2.cuml_params["a"] == 9876
+            dummy2.fit(df, {dummy2.alpha: 9876.0})
```
eordentlich, line 442: Can the test check that model created has alpha=9876.0?

python/src/spark_rapids_ml/params.py
wbo4958: Yeah, super().copy() is a convenient way to avoid handling extra parameters manually.

python/tests/test_common_estimator.py
wbo4958: Done.

wbo4958:
build

python/tests/test_common_estimator.py
```diff
@@ -425,6 +455,31 @@ def assert_model(model: SparkRapidsMLDummyModel) -> None:
...
+            partition_num=ceiling_division(m, max_records_per_batch),
+            m=m,
+            n=n,
+            runtime_check=False,  # don't assert on the runtime.
+        )
+        model3 = dummy3.fit(df, {dummy3.alpha: 9876.0})
+        assert dummy3.cuml_params["a"] == 100
+        assert dummy3.getOrDefault(dummy3.alpha) == 100
+        assert model3.cuml_params["a"] == 9876.0
+        assert model3.getOrDefault(model3.alpha) == 9876.0
```
leewyang, line 481: If `dummy3` is just for testing that `model.alpha == 9876.0`, can we just add the model and asserts to `dummy2` tests above (to avoid another section of tests).

leewyang: Otherwise, LGTM

wbo4958: the `dummy2.fit` will raise an exception when enabling the runtime check, since we're passing alphas=9876 into fit function. In that case, we can't get the model.

python/tests/test_common_estimator.py
eordentlich: Sorry missed the first time, but why should this raise an exception?

wbo4958: the runtime check will enable parameters checking passed from driver side on the executor side.

leewyang (approved):

leewyang (approved):

PR #268, opened 2023-05-22
cast prediction col to double for LinearRegression and RandomForest

wbo4958:
This PR keeps align with the pyspark counterparts for prediction datatype.


wbo4958:
build

PR #273, opened 2023-05-25
fix to gpu assignment under spark-rapids gpu enabled python worker

eordentlich:
spark-rapids python worker daemon sets CUDA_VISIBLE_DEVICES for the launched python worker.   When this env var is set, various CUDA device APIs interpret supplied device indices as indices into the env var value's list vs global gpu ids.    Accordingly, this PR sets the gpu id in this case to 0 as the first entry of the list.

Relevant only for multiple-gpu executors.

### DIFF
--- a/python/src/spark_rapids_ml/utils.py
+++ b/python/src/spark_rapids_ml/utils.py
@@ -92,8 +92,9 @@ def _get_gpu_id(task_context: TaskContext) -> int:
     import os
 
     if "CUDA_VISIBLE_DEVICES" in os.environ and os.environ["CUDA_VISIBLE_DEVICES"]:
+        num_assigned = len(os.environ["CUDA_VISIBLE_DEVICES"].split(","))
         # when CUDA_VISIBLE_DEVICES is set and non-empty, use 0-th index entry
-        return 0
+        gpu_id = 0
     else:
         if task_context is None:
             # safety check.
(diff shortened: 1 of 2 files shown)


python/src/spark_rapids_ml/utils.py
```diff
@@ -89,16 +89,22 @@ def _str_or_numerical(x: str) -> Union[str, float, int]:
...
-        raise RuntimeError(
-            "Couldn't get the gpu id, Please check the GPU resource configuration."
-        )
-    # return the first gpu id.
-    return int(resources["gpu"].addresses[0].strip())
+    import os
+
+    if "CUDA_VISIBLE_DEVICES" in os.environ and os.environ["CUDA_VISIBLE_DEVICES"]:
+        # when CUDA_VISIBLE_DEVICES is set and non-empty, use 0-th index entry
+        return 0
```
leewyang, line 96: Should we warn/log if CUDA_VISIBLE_DEVICES has more than one entry?

eordentlich: We didn't warn before, but it is a good idea.  Will add.

python/src/spark_rapids_ml/utils.py
```diff
@@ -89,16 +89,32 @@ def _str_or_numerical(x: str) -> Union[str, float, int]:
...
-        raise RuntimeError("_get_gpu_id should not be invoked from driver side.")
-    resources = task_context.resources()
-    if "gpu" not in resources:
-        raise RuntimeError(
-            "Couldn't get the gpu id, Please check the GPU resource configuration."
+    import os
+
+    if "CUDA_VISIBLE_DEVICES" in os.environ and os.environ["CUDA_VISIBLE_DEVICES"]:
+        num_assigned = len(os.environ["CUDA_VISIBLE_DEVICES"].split(","))
+        # when CUDA_VISIBLE_DEVICES is set and non-empty, use 0-th index entry
```
wbo4958, line 96: I just tested, if I ran below commands in a bash

``` bash
export CUDA_VISIBLE_DEVICES=""

$ ./deviceQuery 
./deviceQuery Starting...

 CUDA Device Query (Runtime API) version (CUDART static linking)

cudaGetDeviceCount returned 100
-> no CUDA-capable device is detected
Result = FAIL
```

I guess, It will throw an exception if we use gpu_id=0.

eordentlich: Good point.   Ideally we don't get this situation, but we should raise an exception right away in this case instead of simply assigning based on the TaskContext.

eordentlich:
build

lijinf2:
Looks good!

lijinf2 (approved):

leewyang (approved):

leewyang (approved):

PR #281, opened 2023-06-02
BlobsDataGen - Distributed Implementation and Unit Testing

rishic3:
### BlobsDataGen - Implementation + Unit Tests
1. Created gen_data_distributed.py for distributed function implementations.  
   * Created a new base class DataGenBaseMeta with a new function gen_dataframe_and_meta to support returning of meta information, to be used during testing. 
   * Implemented a distributed version of BlobsDataGen; passes unit tests.
2. Created test_gen_data.py to unit test gen_data_distributed functions. Implemented unit testing for BlobsDataGen.
### Misc - gen_data Regression Bug Fix
1. Added "coef" to the set of unsupported parameters for RegressionDataGen in our original gen_data.py. "coef" determines whether coefficients are returned in make_regression, which we currently do not support. Passing "coef" as True would result in a runtime exception.


eordentlich:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,554 @@
...
+    make_blobs,
+    make_classification,
+    make_low_rank_matrix,
+    make_regression,
+)
+
+from benchmark.utils import WithSparkSession, inspect_default_params_from_func, to_bool
+
+
+def dtype_to_pyspark_type(dtype: Union[np.dtype, str]) -> str:
```
lijinf2, line 38: This function seems can be imported from gen_data.py.

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,554 @@
...
+    dtype = np.dtype(dtype)
+    if dtype == np.float32:
+        return "float"
+    elif dtype == np.float64:
+        return "double"
+    else:
+        raise RuntimeError("Unsupported dtype, found ", dtype)
+
+
+class DataGen(object):
```
lijinf2, line 49: Same here. Try to import DataGen class from gen_data.py to avoid duplicate implementations.

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,554 @@
...
+    """DataGen interface"""
+
+    @abstractmethod
+    def gen_dataframe(self, spark: SparkSession) -> Tuple[DataFrame, List[str]]:
+        raise NotImplementedError()
+
+
+class DataGenBase(DataGen):
```
lijinf2, line 57: Same here.  Prefer to import common functions and classes from gen_data.py. If file gen_data.py will be deleted, try to create a new file (e.g. gen_data_utils.py) to host common functions and classes.

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,554 @@
...
+            for k, v in vars(self.args_).items()
+            if k in self.supported_extra_params and v is not None
+        }
+
+    @property
+    def args(self) -> Optional[argparse.Namespace]:
+        return self.args_
+
+
+class DefaultDataGen(DataGenBase):
```
lijinf2, line 179: Seems no need to reimplement this class because it can generate datasets distributedly.

python/benchmark/gen_data_distributed.py
rishic3: OK - I'll import common functions from gen_data for now, and once we've tested and benchmarked both implementations, I can move the definitions over into the gen_data_distributed file. Does that work?

rishic3: Also, I've added the --test argument option in DataGenBase, which wasn't originally in gen_data. OK if I redefine the class in gen_data_distributed?

lijinf2: > OK - I'll import common functions from gen_data for now, and once we've tested and benchmarked both implementations, I can move the definitions over into the gen_data_distributed file. Does that work?

Work for me. Merging can be done later.

lijinf2: > Also, I've added the --test argument option in DataGenBase, which wasn't originally in gen_data. OK if I redefine the class in gen_data_distributed?

Good question. You may redefine by extending the class to create a subclass (e.g. DistributedDataGenBase). 

I feel a better way is to avoid '--test' because it seems an argument used in the test code and it has no effect on user code (i.e. python benchmark/gen_data_distributed.py). You can avoid '--test' by adding an additional variable (e.g. Optional[Dict[str, Any]]) to return values of function DataGenBase::gen_dataframe(..) to return meta information(e.g. centers of BlobsDataGen).

rishic3: With this optional return value, I would still need some way of knowing whether to return the optional argument at all. Wouldn't I still need some flag to indicate that the method call is for testing?

rishic3:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,234 @@
...
+                spark.range(
+                    0, num_partitions, numPartitions=num_partitions
+                ).mapInPandas(make_blobs_udf, schema=",".join(self.schema))
+            ),
+            self.feature_cols,
+            centers,
+        )
+
+
+if __name__ == "__main__":
```
eordentlich, line 146: Looks like a little refactoring would allow avoiding duplication of most of this code from gen_data.py.   ie. move this code to a `main(registered_data_gens)` method in gen_data.py that allows passing in the registered data gens and then invoke (and import in this file) with the different registered_data_gens (original in gen_data.py and distributed in this file).

rishic3: The repartitioning portion is also slightly different in gen_data_distributed (we only repartition for default).

python/benchmark/gen_data_distributed.py
```diff
@@ -0,0 +1,157 @@
+#
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
wbo4958, line 2: The copyright year should be 2023

wbo4958:
Wondering if we need to add data-gen test into the run_test.sh?

wbo4958:
LGTM.

python/benchmark/test_gen_data.py
```diff
@@ -0,0 +1,70 @@
...
+        y = pdf.iloc[:, -1].to_numpy()
+
+        assert X.shape == (50, 2), "X shape mismatch"
+        assert y.shape == (50,), "y shape mismatch"
+        assert centers.shape == (3, 2), "Centers shape mismatch"
+        assert np.unique(y).shape == (3,), "Unexpected number of blobs"
+
+        cluster_stds = [0.7 for _ in range(3)]
+        for i, (ctr, std) in enumerate(zip(centers, cluster_stds)):
+            assert_almost_equal((X[y == i] - ctr).std(), std, 1, "Unexpected std")
```
eordentlich, line 70: Is 1 the tolerance?   Is that a reasonable number given the 0.7 desired?

python/benchmark/test_gen_data.py
```diff
@@ -0,0 +1,70 @@
...
+
+        X = pdf.iloc[:, :-1].to_numpy()
+        y = pdf.iloc[:, -1].to_numpy()
+
+        assert X.shape == (50, 2), "X shape mismatch"
+        assert y.shape == (50,), "y shape mismatch"
+        assert centers.shape == (3, 2), "Centers shape mismatch"
+        assert np.unique(y).shape == (3,), "Unexpected number of blobs"
+
+        cluster_stds = [0.7 for _ in range(3)]
```
eordentlich, line 68: nit .  better as [0.7]*3

python/benchmark/test_gen_data.py
```diff
@@ -0,0 +1,70 @@
...
+        "0.7",
+        "--random_state",
+        "0",
+    ]
+    data_gen = BlobsDataGen(input_args)
+    args = data_gen.args
+    assert args is not None
+    with WithSparkSession(args.spark_confs, shutdown=(not args.no_shutdown)) as spark:
+        df, _, centers = data_gen.gen_dataframe_and_meta(spark)
+        pdf: DataFrame = df.toPandas()
```
eordentlich, line 58: Can you add check on number of partitions?

eordentlich (commented):
Looks good.  A few more minor comments.

python/benchmark/test_gen_data.py
rishic3: 1 here means 1 decimal place precision; same tolerance from SKLearn's unit tests, which I think is reasonable.

python/run_test.sh
```diff
@@ -9,5 +9,6 @@ then
     echo "Tests use at most 4 GPUs. If failed, try setting CUDA_VISIBLE_DEVICES."
 fi
 echo "use --runslow to run all tests"
+pytest -q benchmark/test_gen_data.py
```
eordentlich, line 12: Why `-q` for these tests?

eordentlich:
build

eordentlich:
build

eordentlich (approved):
👍

PR #287, opened 2023-06-06
distributed low rank matrix implementation + unit tests

rishic3:
- Changes: 
   * Implemented a distributed version of LowRankMatrixDataGen. 
   * Added unit testing.
   * Added self to blossom-ci authorization list.


python/benchmark/gen_data_distributed.py
```diff
@@ -144,6 +142,98 @@ def make_blobs_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+        generator = check_random_state(params["random_state"])
+        n = min(rows, cols)
+        # If params not provided, set to defaults.
+        effective_rank = params.get("effective_rank", 10)
+        tail_strength = params.get("tail_strength", 0.5)
+
+        partition_sizes = [rows // num_partitions] * num_partitions
+        partition_sizes[-1] += rows % num_partitions
+        # Check sizes to ensure QR decomp produces a matrix of the correct dimension.
+        for size in partition_sizes:
```
wbo4958, line 190: it seems checking the last two values is enough

rishic3: I think just checking any but the last (e.g. just the first) should be enough, as last is strictly >= non-last

python/benchmark/gen_data_distributed.py
```diff
@@ -144,6 +142,97 @@ def make_blobs_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+        rows = self.num_rows
+        cols = self.num_cols
+        assert self.args is not None
+        num_partitions = self.args.output_num_files
+
+        # Set num_partitions to Spark's default if output_num_files is not provided.
+        if num_partitions is None:
+            num_partitions = spark.sparkContext.defaultParallelism
+
+        generator = check_random_state(params["random_state"])
```
lijinf2, line 181: It seems the purpose is one line code that converts an int to a generator. Maybe there is no need to define a new function check_random_state.

rishic3: It just adds some error checking. We could assume the seed is always a valid int for our purposes

lijinf2:
Would be nice if you can briefly explain the high-level idea.  The PR seems to use SVD and QR to obtain $effective_rank singular vectors, then generate random vectors by calculating different linear combinations of the singular vectors. Not sure if I understand the idea correctly.

rishic3:
> Would be nice if you can briefly explain the high-level idea. The PR seems to use SVD and QR to obtain $effective_rank singular vectors, then generate random vectors by calculating different linear combinations of the singular vectors. Not sure if I understand the idea correctly.

@lijinf2 To make a mostly low rank matrix, we want bell-shaped singular values, where most of the variance can be explained by a bell-shaped curve of width `effective_rank`. 

The low rank part of the singular values profile is: `(1 - tail_strength) * exp(-1.0 * (i / effective_rank) ** 2)`
and the remaining singular values' tail is fat, decreasing as: `tail_strength * exp(-0.1 * i / effective_rank)`

The low rank part can be considered the structured signal, while the tail can be considered the noise that cannot be summarized by the small number of singular components. 

The above describes how we generate S; the QR decomposition is used to generate U and V, the left and right singular vectors of the resultant matrix, of which each contain an orthonormal basis. We use QR decomp (where Q is always orthonormal) with the same seed to ensure U and V are orthonormal. Then we compute our SVD product, yielding a matrix with the desired singular values and orthonormal singular vectors.

python/benchmark/test_gen_data.py
```diff
@@ -63,3 +64,39 @@ def test_make_blobs() -> None:
...
+
+
+def test_make_low_rank_matrix() -> None:
+    input_args = [
+        "--num_rows",
+        "50",
+        "--num_cols",
+        "20",
+        "--dtype",
+        "float64",
```
eordentlich, line 76: Can you parameterize test and sweep over type float64 and float32?   See how parameterization is done in the other tests.

eordentlich: Also, in test blobs.

python/benchmark/gen_data_distributed.py
```diff
@@ -144,6 +142,97 @@ def make_blobs_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+                partition_index = pdf.iloc[0][0]
+                n_partition_rows = partition_sizes[partition_index]
+                u, _ = linalg.qr(
+                    generator.standard_normal(size=(n_partition_rows, n)),
+                    mode="economic",
+                    check_finite=False,
+                )
+                # Include partition-wise normalization to ensure overall unit norm.
+                u *= np.sqrt(1 / num_partitions)
+                mat = np.dot(np.dot(u, s), v.T).astype(dtype)
```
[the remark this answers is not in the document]
rishic3: Not sure if computing the intermediate `math` and then deleting `u` is more memory efficient than yielding `DataFrame(data=np.dot(u, sv_normed))`. I figure since they're the same shape the approaches will be similar

eordentlich:
build

eordentlich:
build

eordentlich (approved):
👍

PR #291, opened 2023-06-08
distributed regression implementation + unit tests

rishic3:
- Changes:
   * Implemented distributed make_regression
   * Implemented unit testing for both default case (X is generated from standard normal) and low rank case (X is generated from low_rank_matrix)


rishic3:
Regression runtime test:
1,000,000 rows, 3000 columns, float32, 50 partitions, 20g driver memory, 16 threads
-> ~9 minutes 5 seconds

rishic3:
build

python/benchmark/test_gen_data.py
```diff
@@ -103,3 +103,91 @@ def test_make_low_rank_matrix(dtype: str) -> None:
...
+    # Effective rank = None
+    input_args = [
+        "--num_rows",
+        "100",
+        "--num_cols",
+        "10",
+        "--dtype",
+        dtype,
+        "--output_dir",
+        "temp",
```
wbo4958, line 164: we have a fixture tmp_path that you can use directly. please refer to https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.06/python/tests/conftest.py#L54-L58

python/benchmark/test_gen_data.py
```diff
@@ -103,3 +103,91 @@ def test_make_low_rank_matrix(dtype: str) -> None:
...
+    # Effective rank = 5
+    input_args = [
+        "--num_rows",
+        "100",
+        "--num_cols",
+        "10",
+        "--dtype",
+        dtype,
+        "--output_dir",
+        "temp",
```
wbo4958, line 119: please use `tmp_path` directly.

python/benchmark/test_gen_data.py
```diff
@@ -103,3 +103,91 @@ def test_make_low_rank_matrix(dtype: str) -> None:
 
         _, s, _ = svd(X)
         assert sum(s) - 5 < 0.1, "X rank is not approximately 5"
+
+
+@pytest.mark.parametrize("dtype", ["float32", "float64"])
+def test_make_regression_low_rank(dtype: str) -> None:
+    # Effective rank = 5
```
wbo4958, line 110: Seems you can combine test_make_regression_low_rank and test_make_regression_well_conditioned with effictive_rank parameterized.

python/benchmark/gen_data_distributed.py
```diff
@@ -232,6 +234,152 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+            # Generate a low-rank, fat tail input set.
+            X, _ = LowRankMatrixDataGen(lrm_input_args).gen_dataframe(spark)
+            assert X.rdd.getNumPartitions() == num_partitions, (
+                f"Unexpected num partitions received from LowRankMatrix;"
+                f"expected {num_partitions}, got {X.rdd.getNumPartitions()}"
+            )
+        else:
+            # Randomly generate a well-conditioned input set.
+            X = spark.createDataFrame(
+                RandomRDDs.normalVectorRDD(
```
wbo4958, line 319: can it generate the regression datasets?

python/benchmark/test_gen_data.py
rishic3: The path isn't actually being used (only provided since it's required) as the test doesn't execute main, just checks the dataframe directly. Should I still bother using the fixture?

python/benchmark/test_gen_data.py
rishic3: To invoke the well-conditioned test, I need to leave off "--effective_rank" entirely; the parser expects either an int or no arg at all. I could combine them into one test, but that would still require defining two sets of input_args and having an if statement.

python/benchmark/gen_data_distributed.py
rishic3: Both generate the regression set; the effective_rank just determines how biased the nput set distribution is.

rishic3:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -232,6 +234,152 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+        noise = params.get("noise", 0.0)
+        shuffle = params.get("shuffle", True)
+        effective_rank = params.get("effective_rank", None)
+        n_informative = params.get("n_informative", 10)
+        n_targets = params.get("n_targets", 1)
+
+        # Input set is either well conditioned (default) or has a low rank fat tail singular profile (see LowRankMatrixDataGen).
+        # Output is generated by applying a (potentially biased) random linear regression model to the input, with n_informative
+        # nonzero regressors and some gaussian centered noise with adjustable scale.
+        # Code adapted from sklearn.datasets.make_regression().
```
lijinf2, line 289: Need to specify "Comments (line 286-288) come from sklearn.datasets.make_regression()." as well.

python/benchmark/gen_data_distributed.py
```diff
@@ -232,6 +234,152 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+                f"expected {num_partitions}, got {X.rdd.getNumPartitions()}"
+            )
+
+        # Generate ground truth upfront.
+        ground_truth = np.zeros((cols, n_targets))
+        ground_truth[:n_informative, :] = 100 * generator.uniform(
+            size=(n_informative, n_targets)
+        )
+
+        if shuffle:
```
lijinf2, line 340: Why shuffle is required in this class?

python/benchmark/gen_data_distributed.py
rishic3: Not sure what you mean - as in, I should specify which exact code lines were adapted from sklearn?

python/benchmark/gen_data_distributed.py
rishic3: Low rank matrix / randomRDDs don't have any shuffling of their own, so row-wise shuffle per partition is meant to reduce some of the potential sample generation bias. Column-wise shuffle is to prevent same indices to be set as informative features every time.

lijinf2:
LGTM

python/benchmark/test_gen_data.py
eordentlich: You can conditionally construct the args list from multiple pieces depending on the value of effective_rank parameter in a parameterized test.

rishic3:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -232,6 +234,155 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+            ]
+            # Generate a low-rank, fat tail input set.
+            X, _ = LowRankMatrixDataGen(lrm_input_args).gen_dataframe(spark)
+            assert X.rdd.getNumPartitions() == num_partitions, (
+                f"Unexpected num partitions received from LowRankMatrix;"
+                f"expected {num_partitions}, got {X.rdd.getNumPartitions()}"
+            )
+        else:
+            # Randomly generate a well-conditioned input set.
+            X = spark.createDataFrame(
```
eordentlich, line 321: Will this be at a different variance than X from low_rank_matrix?

python/benchmark/gen_data_distributed.py
```diff
@@ -232,6 +234,155 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+                X_p = pdf.to_numpy()
+                n_partition_rows = X_p.shape[0]
+
+                if shuffle:
+                    # Column-wise shuffle (global)
+                    X_p[:, :] = X_p[:, indices]
+
+                y = np.dot(X_p, ground_truth) + bias
+                if noise > 0.0:
+                    y += generator.normal(scale=noise, size=y.shape)
```
eordentlich, line 361: So due to the unit column normalization of low rank matrix the 'signal to noise' ratio will depend on the number of examples for a given 'noise' parameter.  Is that how sklearn works?

rishic3: Not sure what you mean here by the number of examples for a given noise parameter.

python/benchmark/gen_data_distributed.py
rishic3: I believe it should be, since the variance from an X in low_rank_matrix will depend on the magnitudes of the singular values, which will vary by the degree of tail_strength.

python/benchmark/gen_data_distributed.py
eordentlich: Will the variance of the predict value decrease with the number examples generated (e.g. due the orthonormality constraint on the output columns in low-rank-matrix)?

python/benchmark/gen_data_distributed.py
eordentlich: Sorry, I meant in terms of how the variance scales/depends on the number of examples generated.

python/benchmark/gen_data_distributed.py
rishic3: Yes, I figure it would. SKLearn's implementation allows this to happen, and they don't ensure any norm on the default cause (with generator.standard_normal) to make the two methods norm-consistent. But in both cases the relative variances of X and y will scale the same, so I imagine it would be fine?

rishic3: And yes, this version is consistent with what sklearn is doing. Normed X with low_rank_matrix, and no norm via the standard normal.

eordentlich:
build

eordentlich:
ci failing formatting

rishic3:
build

rishic3:
build

rishic3:
build

python/benchmark/test_gen_data.py
```diff
@@ -103,3 +103,50 @@ def test_make_low_rank_matrix(dtype: str) -> None:
 
         _, s, _ = svd(X)
         assert sum(s) - 5 < 0.1, "X rank is not approximately 5"
+
+
+@pytest.mark.parametrize(
```
eordentlich, line 108: Missed this before.  Check other examples in the repo on how to set up a multi-parametrized test using multiple parametrize statements.  Don't need to enumerate the combinations as it is done automatically.

rishic3:
build

eordentlich (approved):
👍

PR #296, opened 2023-06-12
distributed make classification + unit testing

rishic3:
runtime test:
1,000,000 rows, 3000 columns, float32, 50 partitions, 20g driver memory, 16 threads
-> ~4 minutes 48 seconds


lijinf2:
> runtime test: 1,000,000 rows, 3000 columns, float32, 50 partitions, 20g driver memory, 16 threads -> ~4 minutes 48 seconds

What is the high-level idea of the implementation in a few sentences?

python/benchmark/test_gen_data.py
```diff
@@ -148,3 +153,60 @@ def test_make_regression(dtype: str, low_rank: bool) -> None:
...
+        "0",
+        "--n_repeated",
+        str(n_repeated),
+        "--hypercube",
+        "True",
+        "--scale",
+        "0.5",
+        "--flip_y",
+        "0",
+        "--random_state",
```
lijinf2, line 187: Is there a way to compare with sklearn make_classification to ensure the set of parameters work as expected?

Seems too much work, can move that in the future.

rishic3: Yeah, their tests are pretty complex and make use of params like weights which we don't support. I tried to pick pieces from their tests that would involve as much coverage as possible.

wbo4958:
Wow, the algo looks so complicated ...

eordentlich:
build

python/benchmark/test_gen_data.py
```diff
@@ -148,3 +153,60 @@ def test_make_regression(dtype: str, low_rank: bool) -> None:
...
+        "--dtype",
+        dtype,
+        "--output_dir",
+        "temp",
+        "--output_num_files",
+        "3",
+        "--n_informative",
+        str(n_informative),
+        "--n_redundant",
+        "0",
```
eordentlich, line 178: Can this be tested on non-zero value with a test for the rank of X?

python/benchmark/test_gen_data.py
```diff
@@ -148,3 +153,60 @@ def test_make_regression(dtype: str, low_rank: bool) -> None:
...
+        assert df.rdd.getNumPartitions() == 3, "Unexpected number of partitions"
+        pdf: DataFrame = df.toPandas()
+        X = pdf.iloc[:, :-1].to_numpy()
+        y = pdf.iloc[:, -1].to_numpy()
+
+        assert X.dtype == np.dtype(dtype), "Unexpected dtype"
+        assert X.shape == (num_rows, 31), "X shape mismatch"
+        assert y.shape == (num_rows,), "y shape mismatch"
+        assert np.unique(y).shape == (2,), "Unexpected number of classes"
+        if num_rows == 2000:
```
eordentlich, line 204: Can you check the 2001 case that it is as expected?

rishic3:
build

eordentlich (approved):
👍

PR #298, opened 2023-06-13
Support regression metrics

wbo4958:
This PR supports regression metrics which has not been applied to other estimators.

### DIFF
--- a/python/src/spark_rapids_ml/metrics/RegressionMetrics.py
+++ b/python/src/spark_rapids_ml/metrics/RegressionMetrics.py
@@ -44,6 +44,47 @@ class _SummarizerBuffer:
         self._weight_square_sum = total_cnt
         self._curr_weight_sum = [total_cnt] * self._num_cols
 
+    def merge(self, other: "_SummarizerBuffer") -> "_SummarizerBuffer":
+        """Merge the other into self and return a new SummarizerBuffer"""
+        self._total_cnt += other._total_cnt
+        self._total_weight_sum += other._total_weight_sum
+        self._weight_square_sum += other._weight_square_sum
+
+        for i in range(self._num_cols):
+            thisWeightSum = self._curr_weight_sum[i]
+            otherWeightSum = other._curr_weight_sum[i]
+            totalWeightSum = thisWeightSum + otherWeightSum
+
+            if totalWeightSum != 0.0:
+                deltaMean = other._curr_mean[i] - self._curr_mean[i]
+                # merge mean together
+                self._curr_mean[i] += deltaMean * otherWeightSum / totalWeightSum
+                # merge m2n together
+                self._curr_m2n[i] += (
+                    other._curr_m2n[i]
+                    + deltaMean
+                    * deltaMean
+                    * thisWeightSum
+                    * otherWeightSum
+                    / totalWeightSum
+                )
+
+            self._curr_weight_sum[i] = totalWeightSum
+            self._curr_m2[i] += other._curr_m2[i]
+            self._curr_l1[i] += other._curr_l1[i]
+
+        return _SummarizerBuffer(
+            self._curr_mean,
+            self._curr_m2n,
+            self._curr_m2,
+            self._curr_l1,
+            self._total_cnt,
+        )
+
+    @property
+    def total_count(self) -> int:
+        return self._total_cnt
+
     @property
     def norm_l2(self) -> DenseVector:
         """L2 (Euclidean) norm of each dimension."""
@@ -104,9 +145,8 @@ class _SummarizerBuffer:
 class RegressionMetrics:
     """Metrics for regression case."""
 
-    def __init__(self, summary: _SummarizerBuffer, throughOrigin: bool = False):
+    def __init__(self, summary: _SummarizerBuffer):
         self._summary = summary
-        self._throughOrigin = throughOrigin
 
     @staticmethod
     def create(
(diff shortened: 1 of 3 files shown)


wbo4958:
build

python/src/spark_rapids_ml/metrics/RegressionMetrics.py
```diff
@@ -0,0 +1,195 @@
...
+        self._curr_mean = mean
+        self._curr_m2n = m2n
+        self._curr_m2 = m2
+        self._curr_l1 = l1
+        self._num_cols = len(mean)
+        # spark-rapids-ml doesn't support weight col, so default to 1 for each sample.
+        self._total_cnt = total_cnt
+        self._total_weight_sum = total_cnt
+        self._weight_square_sum = total_cnt
+        self._curr_weight_sum = [total_cnt] * self._num_cols
```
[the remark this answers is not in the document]
wbo4958: Wow, Really good catch. The weights for label/label-prediction/prediction should be equal. I just followed the Scala version which can be improved.  https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/stat/Summarizer.scala#L441

lijinf2:
> This PR supports regression metrics which has not been applied to other estimators.

Seems the PR introduces a new class RegressionMetrics, and compares the RegressionMetrics with RegressionEvaluator in the test. Can you explain a bit about how RegressionMetrics is different from RegressionEvaluator? Would like to understand when to use RegressionMetrics and not RegressionEvaluator.

python/src/spark_rapids_ml/metrics/RegressionMetrics.py
```diff
@@ -0,0 +1,195 @@
...
+        m2: List[float],
+        l1: List[float],
+        total_cnt: int,
+    ) -> "RegressionMetrics":
+        return RegressionMetrics(_SummarizerBuffer(mean, m2n, m2, l1, total_cnt))
+
+    @property
+    def _ss_y(self) -> float:
+        """sum of squares for label"""
+        return math.pow(self._summary.norm_l2[0], 2)
```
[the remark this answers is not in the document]
wbo4958: Yes. Good catch. Done

python/src/spark_rapids_ml/metrics/RegressionMetrics.py
```diff
@@ -0,0 +1,195 @@
...
+        if denominator > 0.0:
+            delta_mean = self._curr_mean
+            real_variance = [
+                max(
+                    (
+                        self._curr_m2n[i]
+                        + delta_mean[i]
+                        * delta_mean[i]
+                        * self._curr_weight_sum[i]
+                        * (self._total_weight_sum - self._curr_weight_sum[i])
```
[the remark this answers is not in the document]
wbo4958: This part is to calculate the variance, You're right, and this part will always be 0 for our case. I just commented out this.

wbo4958:
> Seems the PR introduces a new class RegressionMetrics, and compares the RegressionMetrics with RegressionEvaluator in the test. Can you explain a bit about how RegressionMetrics is different from RegressionEvaluator? Would like to understand when to use RegressionMetrics and not RegressionEvaluator.

- RegressionEvaluator is the pyspark built-in evaluator which is typically row-wised implemented, so it will not help the performance, especially in the case of CrossValidator. 
- RegressionMetrics is the GPU version metrics for Regression, and it's deeply associated with the specific estimators because we've implemented it in the transformEvaluate.

So, in a word, RegressionMetrics should be hidden and not exposed to users. BTW, we can also implement the GPU version of RegressionEvaluator by RegressionMetrics if needed.

wbo4958:
build

python/tests/test_metrics.py
```diff
@@ -82,3 +84,47 @@ def test_multi_class_metrics(
...
+@pytest.mark.parametrize("metric_name", ["rmse", "mse", "r2", "mae", "var"])
+def test_regression_metrics(metric_name: str) -> None:
+    columns = ["label", "prediction"]
+    np.random.seed(10)
+    pdf1 = pd.DataFrame(
+        np.random.uniform(low=-20, high=20, size=(1000, 2)), columns=columns
+    ).astype(np.float64)
+    np.random.seed(100)
+    pdf2 = pd.DataFrame(
+        np.random.uniform(low=-20, high=20, size=(1000, 2)), columns=columns
```
eordentlich, line 112: Can you make this size different from pdf1?   Say 2000?

wbo4958: Good suggestion, I also added the max_record_batch parameter to test multiple batch scenario.

wbo4958:
build

wbo4958:
@eordentlich could you give another round of review?

python/src/spark_rapids_ml/metrics/RegressionMetrics.py
```diff
@@ -0,0 +1,252 @@
...
+            delta_mean = self._curr_mean
+            real_variance = [
+                max(
+                    (
+                        self._curr_m2n[i]
+                        # + delta_mean[i]
+                        # * delta_mean[i]
+                        # * self._curr_weight_sum[i]
+                        # * (self._total_weight_sum - self._curr_weight_sum[i])
+                        # / self._total_weight_sum
```
eordentlich, line 136: Why leave this commented out code?

eordentlich (approved):
👍

python/src/spark_rapids_ml/metrics/RegressionMetrics.py
wbo4958: Ok, I will delete them in the following PR.

PR #300, opened 2023-06-14
cupy + arrow batching for low_rank_matrix

rishic3:
- Low rank matrix changes:
   * Replaced numpy array operations with cupy
   * Added loop in UDF to yield data in maxRecordsPerBatch chunks for memory efficiency

#### speedups:

3000 cols, float32

- Low rank matrix:

| Samples | Numpy Runtime | Cupy Runtime |
| -------- | -------- | -------- |
| 100,000  | 83s | 43s   |
| 1,000,000  | 603s  | 335s   |

- Regression (low-rank case):

| Samples | Numpy Runtime | Cupy Runtime |
| -------- | -------- | -------- |
| 100,000  | 80s | 67s   |
| 1,000,000  | 675s  | 366s |

If these changes look OK, I'll add arrow batching to the rest of the functions, as well as cupy operations in the regression udf.

### DIFF
--- a/python/benchmark/gen_data_distributed.py
+++ b/python/benchmark/gen_data_distributed.py
@@ -18,8 +18,8 @@ import random
 from abc import abstractmethod
 from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple
 
-import numpy as np
 import cupy as cp
+import numpy as np
 import pandas as pd
 from gen_data import DataGenBase, DefaultDataGen, main
 from pyspark.mllib.random import RandomRDDs
@@ -216,7 +216,9 @@ class LowRankMatrixDataGen(DataGenBase):
         del s
         del v
 
-        maxRecordsPerBatch = spark.sparkContext.getConf().get("spark.sql.execution.arrow.maxRecordsPerBatch")
+        maxRecordsPerBatch = spark.sparkContext.getConf().get(
+            "spark.sql.execution.arrow.maxRecordsPerBatch"
+        )
         if maxRecordsPerBatch is None:
             maxRecordsPerBatch = 10000
         # UDF for distributed generation of U and the resultant product U*S*V.T
@@ -231,7 +233,11 @@ class LowRankMatrixDataGen(DataGenBase):
                 data = cp.dot(u, sv_normed)
                 del u
                 for i in range(0, n_partition_rows, maxRecordsPerBatch):
-                    end_idx = i + maxRecordsPerBatch if i + maxRecordsPerBatch < n_partition_rows else n_partition_rows
+                    end_idx = (
+                        i + maxRecordsPerBatch
+                        if i + maxRecordsPerBatch < n_partition_rows
+                        else n_partition_rows
+                    )
                     yield pd.DataFrame(data=data[i:end_idx].get())
 
         return (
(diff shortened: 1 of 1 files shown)


rishic3:
build

rishic3:
build

wbo4958:
The result looks amazing, but it will introduce a side-effect that the driver must have the cupy installed.

rishic3:
> The result looks amazing, but it will introduce a side-effect that the driver must have the cupy installed.
> 
> Could we move the calculation of sv_normed into numpy on the driver side, and then check how much of the performance we will lose?
> 
> Or move the calculation of sv_normed based on cupy into a single spark job?

After moving sv_normed calculation into numpy on the driver side:

100,000 samples: 43s -> 49s
1,000,000 samples: 335s -> 322s (improved somehow?)

Not a big impact here at all; and in the future I can keep the driver-side free of cupy operations.

lijinf2:
Looks good with Bobby's comment.

In addition to cupy operations, also need to move "import cupy" from driver code to worker code (i.e. make_matrix_udf). In some clusters (e.g. AWS EMR), the driver node may not be equipped with a GPU so "import cupy" may lead to a no-cuda-device error.

python/benchmark/gen_data_distributed.py
```diff
@@ -204,26 +204,42 @@ def gen_dataframe(self, spark: SparkSession) -> Tuple[DataFrame, List[str]]:
...
-                    mode="economic",
-                    check_finite=False,
+                u, _ = cp.linalg.qr(
+                    cp.random.standard_normal(size=(n_partition_rows, n)),
+                    mode="reduced",
                 )
-                yield pd.DataFrame(data=np.dot(u, sv_normed))
+                data = cp.dot(u, cp.asarray(sv_normed))
+                del u
+                for i in range(0, n_partition_rows, maxRecordsPerBatch):
```
eordentlich, line 236: This part will be common to the different generators, I think, so for next one, would be good to find a way to share.

python/benchmark/gen_data_distributed.py
```diff
@@ -204,26 +204,42 @@ def gen_dataframe(self, spark: SparkSession) -> Tuple[DataFrame, List[str]]:
...
+                    end_idx = (
+                        i + maxRecordsPerBatch
+                        if i + maxRecordsPerBatch < n_partition_rows
```
eordentlich, line 239: min(i + maxRecordsBatch, n_partition_rows) ?

rishic3:
build

eordentlich:
Can you rebase to latest branch-23.06 to avoid so many past commits being included in the PR?

wbo4958 (approved):
LGTM

python/benchmark/gen_data_distributed.py
```diff
@@ -204,26 +203,38 @@ def gen_dataframe(self, spark: SparkSession) -> Tuple[DataFrame, List[str]]:
...
-                    check_finite=False,
-                )
-                yield pd.DataFrame(data=np.dot(u, sv_normed))
+                for i in range(0, n_partition_rows, maxRecordsPerBatch):
+                    end_idx = min(i + maxRecordsPerBatch, n_partition_rows)
+                    u, _ = cp.linalg.qr(
+                        cp.random.standard_normal(size=(end_idx - i, n)),
+                        mode="reduced",
+                    )
+                    data = cp.dot(u, cp.asarray(sv_normed))
```
eordentlich, line 235: cp.asarray(sv_normed) seems should be called once outside the loop

rishic3:
Regression (low-rank case): 1,000,000 samples, 3000 features, float32, 20g memory, 75 partitions, 12 threads -> 366 seconds

eordentlich:
build

eordentlich (approved):
👍

PR #315, opened 2023-06-28
Support crossvalidator persistence

wbo4958:
Our customized CrossValidator which reuses the built-in read/write of pyspark CrossValidator has been broken. The root cause is the read process tries to load **spark_rapids_ml.tuning.CrossValidator** from JVM domain because of pyspark hardcoding checking whether it is a python module according to the class name starting with "pyspark.ml" or not. Obviously, our customized CrossValidator starts from "spark_rapids_ml", so the read process tries to load it in JVM domain and finally finds nothing and throws the exception.

### DIFF
--- a/python/src/spark_rapids_ml/tuning.py
+++ b/python/src/spark_rapids_ml/tuning.py
@@ -138,6 +138,8 @@ class CrossValidator(SparkCrossValidator):
 
     @staticmethod
     def _is_python_params_instance(metadata: Dict[str, Any]) -> bool:
+        # If it's not python module, pyspark will load spark_rapids_ml.tuning.CrossValidator
+        # from JVM package. So we need to hack here
         return metadata["class"].startswith(("pyspark.ml.", "spark_rapids_ml."))
 
     @classmethod
--- a/python/src/spark_rapids_ml/tuning.py
+++ b/python/src/spark_rapids_ml/tuning.py
@@ -15,13 +15,14 @@
 #
 
 from multiprocessing.pool import ThreadPool
-from typing import List, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 import numpy as np
 from pyspark import inheritable_thread_target
 from pyspark.ml import Model
 from pyspark.ml.tuning import CrossValidator as SparkCrossValidator
 from pyspark.ml.tuning import CrossValidatorModel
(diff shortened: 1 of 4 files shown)


wbo4958:
build

wbo4958:
build

wbo4958:
@leewyang @eordentlich could you help to review it?

python/src/spark_rapids_ml/tuning.py
```diff
@@ -134,3 +146,32 @@ def singePassTask(
...
+        # If it's not python module, pyspark will load spark_rapids_ml.tuning.CrossValidator
+        # from JVM package. So we need to hack here
+        return metadata["class"].startswith(("pyspark.ml.", "spark_rapids_ml."))
+
+    @classmethod
+    def load(cls, path: str) -> "CrossValidator":
+        orig_is_python_params_instance = DefaultParamsReader.isPythonParamsInstance
+        try:
+            # Replace isPythonParamsInstance
+            setattr(
```
eordentlich, line 161: Can we do this replacement when this package is imported?  And just have it apply to all spark_rapids_ml classes?

python/src/spark_rapids_ml/tuning.py
```diff
@@ -134,3 +146,32 @@ def singePassTask(
...
+                DefaultParamsReader,
+                "isPythonParamsInstance",
+                CrossValidator._is_python_params_instance,
+            )
+            cv_pyspark = super().load(path)
+            cv = cls()
+            cv_pyspark._copyValues(cv)
+        finally:
+            # Must restore to the original isPythonParamsInstance
```
eordentlich, line 170: Why is this necessary?

python/src/spark_rapids_ml/tuning.py
wbo4958: Hmm, I really don't want to break the pyspark things, although we know our changing by replacing the attribute of DefaultParamsReader may not have any impaction on pyspark.

python/src/spark_rapids_ml/tuning.py
wbo4958: Yeah, From the function we customized, we don't need to restore to the original function. But again, I really don't want to hack pyspark internal things. To be honest, I don't know if this hacked function will have other impacts on the pyspark.

eordentlich (approved):
👍

PR #319, opened 2023-07-03
Support gpu version crossvalidator for LinearRegression

wbo4958:


wbo4958:
build

python/src/spark_rapids_ml/regression.py
```diff
@@ -524,25 +608,30 @@ def _create_pyspark_model(self, result: Row) -> "LinearRegressionModel":
...
+    _RegressionModelEvaluationMixIn,
 ):
     """Model fitted by :class:`LinearRegression`."""
 
     def __init__(
         self,
-        coef_: List[float],
-        intercept_: float,
+        coef_: Union[List[float], List[List[float]]],
+        intercept_: Union[float, List[float]],
```
leewyang, line 626: Wondering if just assuming lists here would simplify things elsewhere, where the single model is just represented as a list of one item.

leewyang (approved):
One question/comment, but otherwise LGTM

python/src/spark_rapids_ml/regression.py
wbo4958: Wow, Good suggestion. I will file a followup PR to handle it not only for LinearRegression but also for RandomForest.

PR #321, opened 2023-07-06
umap fit() implementation and tests

rishic3:
- future additions:
   * OOM detection of data subsample (requires running a mini spark job to query GPU memory on node)
   * "convert_dtype" fit() param support


python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+            "repulsion_strength": 1.0,
+            "negative_sample_rate": 5,
+            "transform_queue_size": 4.0,
+            "a": None,
+            "b": None,
+            "hash_input": False,
+            "random_state": None,
+            "callback": None,
+            "verbose": False,
+            "output_type": None,
```
wbo4958, line 80: Seems we don't need handle and output_type as the parameter.

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+            "set_op_mix_ratio": 1.0,
+            "local_connectivity": 1.0,
+            "repulsion_strength": 1.0,
+            "negative_sample_rate": 5,
+            "transform_queue_size": 4.0,
+            "a": None,
+            "b": None,
+            "hash_input": False,
+            "random_state": None,
+            "callback": None,
```
wbo4958, line 78: could you please double check if spark can support callback?

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+    def setLabelCol(self: P, value: str) -> P:
+        """
+        Sets the value of :py:attr:`labelCol`.
+        """
+        return self.set_params(labelCol=value)
+
+
+class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
+
+    """
```
wbo4958, line 126: Please add more doc for UMAP.

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+    >>> from spark_rapids_ml.umap import UMAP
+    >>> X, _ = (1000, 10, centers=42, cluster_std=0.1,
```
wbo4958, line 128: seems missing a function?

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+                dtype=np.float32, random_state=10)
+    >>> df = spark.createDataFrame(X, ["features"])
+    >>> local_model = UMAP().setFeaturesCol("features")
+    >>> distributed_model = umap.fit(df)
+    >>> embeddings = distributed_model.transform(df)
+    >>> embeddings.show()
+    """
+
+    def __init__(self, sample_fraction: float = 1.0, **kwargs: Any) -> None:
```
wbo4958, line 137: please expand the kwargs to the specific parameters, so we can have nice documentaion

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,280 @@
...
+                seed=self.cuml_params["random_state"],
+            )
+        else:
+            data_subset = dataset
+
+        input_num_workers = self.num_workers
+        # Force to single partition, single worker
+        self._num_workers = 1
+        if data_subset.rdd.getNumPartitions() != 1:
+            data_subset = data_subset.coalesce(1)
```
wbo4958, line 159: Have you tried this case? I'm not quite sure if coalesce can work with the following barrier RDD.

python/tests/test_umap.py
```diff
@@ -0,0 +1,165 @@
...
+        n_parts,
+        n_rows,
+        sampling_ratio,
+        supervised,
+        dataset,
+        n_neighbors,
+        dtype,
+    )
+
+    if not result:
```
wbo4958, line 154: Could you please explain why we need to rerun it? Isn't the result determinate?

python/tests/test_umap.py
```diff
@@ -0,0 +1,165 @@
...
+    with CleanSparkSession() as spark:
+        if supervised:
+            data_df, features_col, label_col = create_pyspark_dataframe(
+                spark, "array", dtype, local_X, local_y
+            )
+            assert label_col is not None
+            local_model.setLabelCol(label_col)
+        else:
+            data_df, features_col, _ = create_pyspark_dataframe(
+                spark, "array", dtype, local_X, None
```
wbo4958, line 93: Since there is only 1 test, I'd like to suggest that we'd better cover more different spark types, like vector/columnar input

wbo4958 (commented):
It would be great if another test can be added to test the UMAP estimator persistence.

python/tests/test_umap.py
rishic3: This repetition is copied from cuml - you're right that the fit() should be deterministic, but this is primarily for testing transform(), which has stochasticity.

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,426 @@
...
+                # Call unsupervised fit
+                local_model = umap_object.fit(concated)
+
+            dtype = str(local_model.embedding_.dtype)
+
+            return {
+                "embedding_": [local_model.embedding_.tolist()],
+                "raw_data_": [concated.tolist()],
+                "n_cols": len(local_model.embedding_[0]),
+                "dtype": dtype,
```
eordentlich, line 340: We could consider return the embeddings and raw data in smaller chunks (e.g. based on arrow batch size parameter) with one embedding/data row per row of returned df.   Not sure if having corresponding embedding and data rows in the same pdf row (i.e. two columns) is better than first all embedding rows and then all data rows say, in a single column.   Former seems logically cleaner.

Can n_cols and dtype be determined on driver?

rishic3:
#### Idea to avoid overriding _call_cuml_fit_func():
Add class method `def _per_row_fit_return(self) -> bool` (similar to _fit_array_order or _require_nccl_ucx) that determines whether to do a row-wise return from call_cuml_fit_func(). Add conditional within core.__call_cuml_fit_func() to optionally return fit data row by row as a bonafide RDD based on this class flag.

python/tests/test_umap.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+
+    loc_umap = _local_umap_trustworthiness(local_X, local_y, n_neighbors, supervised)
+
+    print("Local UMAP trustworthiness score : {:.2f}".format(loc_umap))
+    print("Spark UMAP trustworthiness score : {:.2f}".format(dist_umap))
+
+    trust_diff = loc_umap - dist_umap
+
+    return trust_diff <= 0.15
```
lijinf2, line 143: Is there any intuition to set the tolerance to 0.15 and not other value?  Will be good to add a comment to explain.

rishic3: This is taken from the cuml unit tests, so hopefully it is based on some empirical testing on their part. The trustworthiness metric ranges from 0-1, and given the stochasticity of transform, some lenience makes sense. Not sure if 0.15 is the ideal value though, they don't have any written justification.

python/src/spark_rapids_ml/umap.py
wbo4958: I mean can we add the specific function parameters into `__init__` function

``` python
def __init__(self, n_neighbors: int = 25, n_components: int = 2, ...)
```

wbo4958:
> 



> #### Idea to avoid overriding _call_cuml_fit_func():
> Currently, override of _call_cuml_fit_func is only necessitated by the yield statement.
> 
> Alternatively, add class method `def _per_row_fit_return(self) -> bool` (similar to _fit_array_order or _require_nccl_ucx). Add conditional within core._call_cuml_fit_func() to optionally return data row by row as a bonafide RDD based on this class flag. (may be useful if future algos need this too.)
> 
> Open to other (cleaner?) suggestions.

I would suggest the 2nd way which can re-use code.

lijinf2:
> #### Idea to avoid overriding _call_cuml_fit_func():
> Currently, override of _call_cuml_fit_func is only necessitated by the yield statement.
> 
> Alternatively, add class method `def _per_row_fit_return(self) -> bool` (similar to _fit_array_order or _require_nccl_ucx). Add conditional within core._call_cuml_fit_func() to optionally return data row by row as a bonafide RDD based on this class flag. (may be useful if future algos need this too.)
> 
> Open to other (cleaner?) suggestions.

Is it possible to set pyspark configuration to increase this serialization limit? One way is to check the expected serialization limit and hint user to increase the pyspark configuration parameter. 

another option is to return row by row in core.py for all algorithms. (any downside?)

rishic3:
> > #### Idea to avoid overriding _call_cuml_fit_func():
> > Currently, override of _call_cuml_fit_func is only necessitated by the yield statement.
> > Alternatively, add class method `def _per_row_fit_return(self) -> bool` (similar to _fit_array_order or _require_nccl_ucx). Add conditional within core._call_cuml_fit_func() to optionally return data row by row as a bonafide RDD based on this class flag. (may be useful if future algos need this too.)
> > Open to other (cleaner?) suggestions.
> 
> Is it possible to set pyspark configuration to increase this serialization limit? One way is to check the expected serialization limit and hint user to increase the pyspark configuration parameter.
> 
> another option is to return row by row in core.py for all algorithms. (any downside?)

There is the driver.maxResultSize that can be set to unlimited, but I don't think there's a way to increase the JVM 2GB byte array limit. 

As for returning row-by-row for all algos, I don't think this would work - UMAP is not implementing the typical get_cuml_fit_func and instead using a generator function, which is why we can return row-by-row during call_cuml_fit, so we would have to also make this change for other algos. Probably adding a conditional withi call_cuml_fit is best

rishic3:
Avoiding override of call_cuml_fit_func by using a _fit_return_by_row() class flag to determine which fit_func to use (generator or not) and how to return data. Does require a `#type: ignore` though, since the fit_generator_func may not be implemented.

python/src/spark_rapids_ml/core.py
```diff
@@ -532,6 +560,11 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
                 result = cuml_fit_func(inputs, params)
                 logger.info("Cuml fit complete")
 
+                if use_generator:
```
eordentlich, line 563: Should this be in the CumlContext?

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,476 @@
...
+                    labels = _concat_and_free(label_list, order=array_order)
+                local_model = umap_object.fit(concated, y=labels)
+            else:
+                # Call unsupervised fit
+                local_model = umap_object.fit(concated)
+
+            embedding = local_model.embedding_
+            del local_model
+
+            for embedding, raw_data in zip(embedding, concated):
```
eordentlich, line 379: Is this not too slow to create a DataFrame for each row?   Does working in chunks and yielding List[List[]] for each chunk in each corresponding "Row" work?

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,476 @@
...
+        # Collect and concatenate row-by-row fit results
+        embeddings = [row["embedding_"] for row in rows]
+        raw_data = [row["raw_data_"] for row in rows]
+        del rows
+
+        model = UMAPModel(
+            embedding_=embeddings,
+            raw_data_=raw_data,
+            n_cols=len(raw_data[0]),
+            dtype=str(np.array(raw_data[0][0]).dtype),
```
eordentlich, line 319: Would this dtype just be float64 always?

python/tests/test_umap.py
```diff
@@ -0,0 +1,270 @@
...
+    local_y: np.ndarray,
+    n_neighbors: int,
+    supervised: bool,
+    n_parts: int,
+    n_workers: int,
+    sampling_ratio: float,
+    dtype: np.dtype,
+    feature_type: str,
+) -> float:
+    local_model = UMAP(
```
eordentlich, line 84: Here as well, wouldn't this be better called an estimator vs local_model?

eordentlich (commented):
some additional comments.   looks good overall.

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,476 @@
...
+        )
+        model._num_workers = input_num_workers
+
+        self._copyValues(model)
+        self._copy_cuml_params(model)  # type: ignore
+
+        return model
+
+    def _fit_array_order(self) -> _ArrayOrder:
+        return "C"
```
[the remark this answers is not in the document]
rishic3: Yup, cuml uses C ordering when creating their arrays.

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,476 @@
...
+                local_model = umap_object.fit(concated)
+
+            embedding = local_model.embedding_
+            del local_model
+
+            for embedding, raw_data in zip(embedding, concated):
+                yield pd.DataFrame(
+                    data=[
+                        {
+                            "embedding_": embedding.tolist(),
```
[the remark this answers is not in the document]
rishic3: It will convert to python float, and the lists returned are also python float, so I think it's actually not an issue (remains single precision).

python/src/spark_rapids_ml/umap.py
rishic3: I'll fix this. I should just check the python type at this point; getting the numpy version of a python float seems to just default to float64.

rishic3:
build

python/src/spark_rapids_ml/umap.py
```diff
@@ -0,0 +1,524 @@
...
+
+        embeddings = list(chain.from_iterable([row["embedding_"] for row in rows]))
+        raw_data = list(chain.from_iterable([row["raw_data_"] for row in rows]))
+        del rows
+
+        model = UMAPModel(
+            embedding_=embeddings,
+            raw_data_=raw_data,
+            n_cols=len(raw_data[0]),
+            dtype="np.float32" if isinstance(raw_data[0][0], float) else "np.float64",
```
eordentlich, line 361: Looks like python float is generally double precision.   No single precision in python (def. on my linux ws, but probably vast majority of systems):  Can try this https://docs.python.org/3.10/library/sys.html#sys.float_info

rishic3: Was just playing with this in a notebook. No matter the decimal places provided, python will allocate double-precision space for the floats. As long as we're returning lists, don't think there's a way to avoid it.

rishic3:
build

eordentlich (approved):
LGTM

PR #326, opened 2023-07-11
remove some print statements

eordentlich:
Didn't catch these before.   The second one was printing a lot for our benchmark script.


eordentlich:
build

leewyang (approved):

PR #330, opened 2023-07-22
benchmarking related updates

eordentlich:
mostly to bring local script more inline with cloud versions and add parameterization via env vars.

### DIFF
--- a/python/benchmark/gen_data_distributed.py
+++ b/python/benchmark/gen_data_distributed.py
@@ -85,6 +85,8 @@ class BlobsDataGen(DataGenBaseMeta):
 
         print(f"Passing {params} to make_blobs")
 
+        include_labels = params.pop("include_labels", False)
+
         rows = self.num_rows
         cols = self.num_cols
         assert self.args is not None
@@ -104,14 +106,12 @@ class BlobsDataGen(DataGenBaseMeta):
 
         # Generate centers upfront.
         _, _, centers = make_blobs(
-            n_samples=0, n_features=cols, **params, return_centers=True
+            n_samples=1, n_features=cols, **params, return_centers=True
         )
 
         # Update params for partition-specific calls.
         params["centers"] = centers
         del params["random_state"]
-
-        include_labels = params.get("include_labels", False)
 
         maxRecordsPerBatch = int(
             spark.sparkContext.getConf().get(
@@ -382,7 +382,9 @@ class RegressionDataGen(DataGenBaseMeta):
                     cols,
                     numPartitions=num_partitions,
                     seed=seed,
-                ).map(lambda nparray: nparray.tolist()),
+                ).map(
+                    lambda nparray: nparray.tolist()  # type: ignore
+                ),
                 schema=",".join(self.schema),
             )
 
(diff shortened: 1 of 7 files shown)


python/benchmark/gen_data_distributed.py
```diff
@@ -103,7 +106,7 @@ def gen_dataframe_and_meta(
 
         # Generate centers upfront.
         _, _, centers = make_blobs(
-            n_samples=0, n_features=cols, **params, return_centers=True
+            n_samples=1, n_features=cols, **params, return_centers=True
```
eordentlich, line 109: latest version of sklearn complained about this being 0

python/benchmark/gen_data_distributed.py
```diff
@@ -375,7 +382,9 @@ def gen_dataframe_and_meta(
                     cols,
                     numPartitions=num_partitions,
                     seed=seed,
-                ).map(lambda nparray: nparray.tolist()),
+                ).map(
+                    lambda nparray: nparray.tolist()  # type: ignore
```
eordentlich, line 386: Not sure about this, but I think under spark 3.3.1, this type checking fails.

eordentlich:
build

python/run_benchmark.sh
```diff
@@ -10,115 +10,273 @@
...
+# if num_cols=3000 => arrow_batch_size=20000, scale linearly for smaller number of columns
+arrow_batch_size=$(( 20000 * ( ( $num_cols + 3000 - 1 ) / $num_cols ) ))
+
 
 # stop on first fail
 set -e
 
+sep="=================="
+
+# TODO adjust maxRecorsPerBatch based on num columns
```
leewyang, line 85: Is this comment still relevant?  (If so, there's a typo).

python/run_benchmark.sh
```diff
@@ -10,115 +10,273 @@
...
-        --num_cpus 0 \
-        --train_path "/tmp/blobs/5k_3k_float32.parquet" \
-        --report_path "report_kmeans.csv" \
-        --spark_confs "spark.master=local[4]" \
-        --spark_confs "spark.driver.memory=128g" \
-        --spark_confs "spark.sql.execution.arrow.maxRecordsPerBatch=20000" \
+        --k 1000 \
+        --tol 1.0e-20 \
+        --maxIter 30 \
+        --initMode random \
```
leewyang, line 157: Any specific reason why these param values were chosen?  i.e. should we document that reason here?  Same with all other algos.

leewyang (approved):
Minor comments, otherwise LGTM

python/run_benchmark.sh
eordentlich: The args now match what was used in our blog post and the cloud benchmarks.   High compute regime.

python/run_benchmark.sh
eordentlich: No longer relevant.  Deleted.

eordentlich:
build

leewyang (approved):

wbo4958 (approved):

PR #331, opened 2023-07-22
fix and update pip dockerfile

eordentlich:
rapids no longer supports python 3.8 which is base ubuntu 20.04 images.  
updating to 22.04 brings in python 3.10.

enable building on arm systems as well (some pip packages require compilation during install - e.g. treelite).


leewyang:
build

leewyang:
build

leewyang (approved):

PR #332, opened 2023-07-22
umap transform() and tests

rishic3:


python/src/spark_rapids_ml/umap.py
```diff
@@ -507,19 +507,68 @@ def raw_data(self) -> List[List[float]]:
...
         return (False, False)
 
     def _out_schema(self, input_schema: StructType) -> Union[StructType, str]:
-        raise NotImplementedError("TODO")
+        return StructType(
+            [
+                StructField("input", ArrayType(FloatType(), False), False),
+                StructField("embedding", ArrayType(FloatType(), False), False),
+            ]
+        )
```
leewyang, line 574: Remove "input" (if not necessary per above).  Also, avoid hard-coding "embedding" and use the value of `getOutputCol()` or `getPredictionCol()`  instead (which can be set to a default value of "embedding", if desired).

python/src/spark_rapids_ml/umap.py
```diff
@@ -507,19 +507,68 @@ def raw_data(self) -> List[List[float]]:
...
+                embedding[i, :] if is_emb_np else embedding.iloc[i, :]
+                for i in range(embedding.shape[0])
+            ]
+
+            result = pd.DataFrame(
+                {
+                    "input": input_list,
+                    "embedding": emb_list,
+                }
+            )
```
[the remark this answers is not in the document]
rishic3: It's being used for the trustworthiness test, which requires the sample order to be preserved

python/src/spark_rapids_ml/umap.py
```diff
@@ -139,6 +139,21 @@ def setLabelCol(self: P, value: str) -> P:
         Sets the value of :py:attr:`labelCol`.
         """
         return self.set_params(labelCol=value)
+    
+    def getOutputCol(self) -> str:
+        """
+        Gets the value of :py:attr:`outputCol`. Contains the embeddings of the input data.
+        """
+        if self.isDefined(self.outputCol):
+            return self.getOrDefault("outputCol")
```
wbo4958, line 148: Seems no need to check whether the outputCol has been configured. It's ok to just return self.getOrDefault("outputCol")

python/src/spark_rapids_ml/umap.py
```diff
@@ -315,6 +333,7 @@ def __init__(self, sample_fraction: float = 1.0, **kwargs: Any) -> None:
         self.set_params(**kwargs)
         self.sample_fraction = sample_fraction
         self.maxRecordsPerBatch = 10000
```
wbo4958, line 335: Is there any reason to hardcode the maxRecordsPerBatch to 10000?

python/src/spark_rapids_ml/umap.py
```diff
@@ -507,19 +526,67 @@ def raw_data(self) -> List[List[float]]:
     def _get_cuml_transform_func(
         self, dataset: DataFrame, category: str = transform_evaluate.transform
     ) -> Tuple[_ConstructFunc, _TransformFunc, Optional[_EvaluateFunc],]:
+        cuml_alg_params = self.cuml_params.copy()
```
wbo4958, line 529: seems "copy" is not needed here.

python/src/spark_rapids_ml/umap.py
```diff
@@ -507,19 +526,67 @@ def raw_data(self) -> List[List[float]]:
...
+        driver_embedding = np.array(self.embedding_, dtype=np.float32)
```
wbo4958, line 530: maybe the broadcast way can be more efficient since we don't need to package the embedding or raw_data to every task and send it to executor from driver

python/src/spark_rapids_ml/umap.py
```diff
@@ -507,19 +526,67 @@ def raw_data(self) -> List[List[float]]:
...
+        driver_raw_data = np.array(self.raw_data_, dtype=np.float32)
+
         def _construct_umap() -> CumlT:
-            raise NotImplementedError("TODO")
+            import cupy as cp
+            from cuml.common import SparseCumlArray, input_to_cuml_array
```
wbo4958, line 535: please use utils.cudf_to_cuml_array since there is a compatibility issue with the older cuml

python/src/spark_rapids_ml/umap.py
rishic3: Just a placeholder to avoid type-check issues, since it gets set again later in _fit(). 10,000 is the default from pyspark.sql.execution.arrow.

leewyang:
build

python/tests/test_umap.py
```diff
@@ -147,9 +149,7 @@ def _run_spark_test(
 @pytest.mark.parametrize("n_parts", [2, 9])
 @pytest.mark.parametrize("n_workers", [12])
 @pytest.mark.parametrize("n_rows", [100, 500])
-@pytest.mark.parametrize(
-    "sampling_ratio", [1.0]
-)  # Temporarily set to full dataset for fit() testing
+@pytest.mark.parametrize("sampling_ratio", [0.55, 0.9])
 @pytest.mark.parametrize("supervised", [True, False])
 @pytest.mark.parametrize("dataset", ["digits", "iris"])
 @pytest.mark.parametrize("n_neighbors", [10])
```
leewyang, line 155: Please add `@pytest.mark.slow` annotation to this `test_spark_umap` parametrize test, since this test takes a long time. (See other algos for example).

leewyang: Might also want to add a simple smoke test for normal "non-slow" testing.

rishic3:
build

leewyang (approved):

PR #344, opened 2023-07-26
Support stage level scheduling

wbo4958:
The precondition that stage-level scheduling can work in the spark-rapids-ml is the executors have already grabbed the GPU resource at every beginning even for the ETL phase. The stage-level scheduling in spark-rapids-ml won’t require extra executors, instead, it just requires existing different resources for different stages.

### DIFF
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -699,7 +699,8 @@ class _CumlEstimator(Estimator, _CumlCaller):
         # each training task requires cpu cores > total executor cores/2 which can
         # ensure each training task be sent to different executor.
         task_cores = (int(executor_cores) // 2) + 1
-        task_gpus = 0.1 if task_gpu_amount is None else float(task_gpu_amount)
+        # To ensure the training task take the whole GPU exclusively
+        task_gpus = 1.0
 
         treqs = TaskResourceRequests().cpus(task_cores).resource("gpu", task_gpus)
         rp = ResourceProfileBuilder().require(treqs).build
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -696,10 +696,13 @@ class _CumlEstimator(Estimator, _CumlCaller):
         from pyspark.resource.profile import ResourceProfileBuilder
         from pyspark.resource.requests import TaskResourceRequests
 
-        # Each training task requires only 1 cpu cores to allow other cpu tasks
-        # running alongside the training tasks
-        # and requires 1 gpu to ensure the training task be sent to different executors
-        treqs = TaskResourceRequests().cpus(1).resource("gpu", 1.0)
+        # each training task requires cpu cores > total executor cores/2 which can
+        # ensure each training task be sent to different executor.
+        task_cores = (int(executor_cores) // 2) + 1
+        # To ensure the training task take the whole GPU exclusively
+        task_gpus = 1.0
+
+        treqs = TaskResourceRequests().cpus(task_cores).resource("gpu", task_gpus)
         rp = ResourceProfileBuilder().require(treqs).build
 
         return rdd.withResources(rp)
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -661,10 +661,16 @@ class _CumlEstimator(Estimator, _CumlCaller):
             return super().fitMultiple(dataset, paramMaps)
 
     def _try_stage_level_scheduling(self, rdd: RDD) -> RDD:
+        sc = _get_spark_session().sparkContext
         executor_cores = (
-            _get_spark_session().sparkContext.getConf().get("spark.executor.cores")
+            sc.getConf().get("spark.executor.cores")
         )
-        if executor_cores is None:
+        if _is_local(sc):
+            # TODO verify if stage level scheduling works on local mode.
+            # Local mode doesn't support stage level scheduling
+            return rdd
+        elif executor_cores is None:
+            # TODO to check if there is other way to get executor cores.
             self.logger.warning(
                 "Stage level scheduling in XGBoost "
                 "requires spark.executor.cores to be set "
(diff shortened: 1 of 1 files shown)


wbo4958:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -660,6 +660,52 @@ def fitMultipleModels() -> List["_CumlModel"]:
...
+            )
+            return rdd
+
+        from pyspark.resource.profile import ResourceProfileBuilder
+        from pyspark.resource.requests import TaskResourceRequests
+
+        # each training task requires cpu cores > total executor cores/2 which can
+        # ensure each training task be sent to different executor.
+        task_cores = (int(executor_cores) // 2) + 1
+        task_gpus = 0.1 if task_gpu_amount is None else float(task_gpu_amount)
```
eordentlich, line 702: Would it suffice to set task_gpus = executor_gpu_amount?

wbo4958: Hmm, `task_gpus = executor_gpu_amount` is going to limit only the training task running. If we set it to a fraction, we can still allow other ETL tasks to run alongside training tasks.

eordentlich: You mean from other Spark jobs running in parallel?   I suppose it could also allow other tasks on the GPU as well.   Do we want to allow sharing of GPU with another Spark job?   Seems a recipe for GPU OOMs.

python/src/spark_rapids_ml/core.py
```diff
@@ -660,6 +660,52 @@ def fitMultipleModels() -> List["_CumlModel"]:
...
+                f"the number of runnable tasks per executor to: 1. Please adjust your configuration."
+            )
+            return rdd
+
+        from pyspark.resource.profile import ResourceProfileBuilder
+        from pyspark.resource.requests import TaskResourceRequests
+
+        # each training task requires cpu cores > total executor cores/2 which can
+        # ensure each training task be sent to different executor.
+        task_cores = (int(executor_cores) // 2) + 1
```
lijinf2, line 701: Does this limit the number of tasks per executor to 1? 
Not quite understand the comment in line 700. Seems it indicates one executor one task.

lijinf2: If an application (e.g. PCA) specifies "spark.executor.resource.gpu.amount=4", would line 701 restrict the number of GPU in use to be 1 per executor, while the application expects 4 per executor?

python/src/spark_rapids_ml/core.py
wbo4958: Really good point. so do you suggest setting task_gpus = executor_gpu_amount? @eordentlich

eordentlich: Would task_gpus=1  actually be the ideal choice for now?  That would address @lijinf2 's concern as well.   
But if improved performance can be demonstrated for some use case involving parallel cpu jobs with gpu jobs, maybe we should allow it (and just resolve OOMs via resource configs).
Also, the fit stage scheduling probably should not lower task_gpus below the original setting so that it can be pinned to 1 if needed.

wbo4958: Good. new commit has changed it.

wbo4958:
build

wbo4958:
build

eordentlich (approved):
Nice.  Let's try it in our benchmarks.

python/src/spark_rapids_ml/core.py
lijinf2: Thanks for setting task_gpus to 1. Seems also need to set task_cores at line 701 to 1, so that executor.gpu.amount=4 tasks can be launched in the same executor. The current "task_cores = (int(executor_cores) // 2) + 1" restricts the #task to be 1 per executor.

eordentlich (changes requested):
Forgot about @lijinf2 issue when I approved.   Would be good to make that change.

python/src/spark_rapids_ml/core.py
wbo4958: yes. good catch. thx

wbo4958: Oops, No, if we change task_cores=1, task_gpus=1.0, the stage-level scheduling seems won't work. I just reverted the change.

lijinf2:
Nice. Can't wait to try.

lijinf2:
build

lijinf2 (approved):

eordentlich (approved):

wbo4958:
build

python/src/spark_rapids_ml/core.py
wbo4958: Finally, I got it, task_gpus means how many gpu slots the task requires, but it doesn't mean how many GPU fractions it will use. So we must use task_cpu to limit the concurrent tasks.

leewyang (approved):

python/src/spark_rapids_ml/core.py
eordentlich: @wbo4958 What is the difference between a gpu slot and GPU fraction?   What is the meaning of gpu slot, in this context?

wbo4958: GPU slots are calculated according to the ResourceProfile0 (1 / task.gpu.amount), So if the task.gpu.amount is set to fraction, there'll be more than 1 GPU slot possible. While in stage-level scheduling,  "task.gpu.amount" task requires is how many gpu slots the task needs, if the value is 0~0.5, or 1, it means the task requires only 1 GPU slot,

eordentlich: So if task_gpus amount in stage level scheduling is set to 1/task.gpu.amount would that also be a way to limit to 1 task per gpu for that stage?

wbo4958: Not exactly, the task_gpus amount in stage-level scheduling is in 0~0.5 or 1.0, which means the task requires only 1 GPU slot. Let's assume the spark.task.resource.gpu.amount = 0.2, so there will be 5 GPU slots, In that case, even task_gpu account in stage-level is 1.0, it just requires only 1 slot instead of the whole GPU (5 slots), so the left 4 slots still can be offered to other tasks. 

In that case, we can't use task_gpu amount to limit the tasks.

PR #352, opened 2023-08-07
Override call_cuml_fit_func to use Dataframe, model saving+loading as numpy

rishic3:
### override call_cuml_fit_func
- removed barrier stages, returns dataframe rather than rdd
- allows for:
   * coalesce() rather than repartition
   * toPandas() rather than collect
   * storing raw_data / embeddings as numpy rather than list
   * saving attributes as float32 rather than python float (float64)
- speedups below

dataset: **50,000 x 3000, float32, parquet**
| | fit runtime | transform runtime |
| -------- | -------- | -------- |
| rdd + repartition + collect (no overrides) | 58.5s | 29.2s |
| df + coalesce + collect (python lists) | 46.7s | 28.9s |
| df + coalesce + toPandas (numpy arrays) | 24.9s | 8.7s |

dataset: **100,000 x 3000, float32, parquet**
| | fit runtime | transform runtime |
| -------- | -------- | -------- |
| rdd + repartition + collect (no overrides) | 113.3s | 65.4s |
| df + coalesce + toPandas (numpy arrays) | 48.5s | 23.7s |

### override modelwriter / modelreader
- subclassed cumlmodelreader and cumlmodelwriter to handle numpy saving + loading
   * saves arrays with np.save, creates subdirectory for other model attributes ("metadata")
- allows for continuous use of numpy arrays between fit and transform phases
- saves memory and preserves float32 dtype

### DIFF
--- a/python/src/spark_rapids_ml/umap.py
+++ b/python/src/spark_rapids_ml/umap.py
@@ -409,10 +409,6 @@ class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
             dtype=type(raw_data[0][0]).__name__,
         )
 
-        print(
-            type(raw_data[0][0]).__name__,
-        )
-
         model._num_workers = input_num_workers
 
         self._copyValues(model)
@@ -656,14 +652,14 @@ class UMAPModel(_CumlModel, UMAPClass, _UMAPCumlParams):
     @property
     def embedding(self) -> np.ndarray:
         if isinstance(self.embedding_, np.ndarray):
-            return self.embedding_
-        return self.embedding_.value
+            return self.embedding_.tolist()
+        return self.embedding_.value.tolist()
 
     @property
     def raw_data(self) -> np.ndarray:
         if isinstance(self.raw_data_, np.ndarray):
-            return self.raw_data_
-        return self.raw_data_.value
+            return self.raw_data_.tolist()
+        return self.raw_data_.value.tolist()
 
     def _get_cuml_transform_func(
         self, dataset: DataFrame, category: str = transform_evaluate.transform
@@ -678,8 +674,17 @@ class UMAPModel(_CumlModel, UMAPClass, _UMAPCumlParams):
 
             from .utils import cudf_to_cuml_array
 
-            embedding = self.embedding
-            raw_data = self.raw_data
+            embedding = (
+                self.embedding_
+                if isinstance(self.embedding_, np.ndarray)
+                else self.embedding_.value
+            )
+            raw_data = (
+                self.raw_data_
+                if isinstance(self.raw_data_, np.ndarray)
+                else self.raw_data_.value
+            )
+
             if embedding.dtype != np.float32:
                 embedding = embedding.astype(np.float32)
                 raw_data = raw_data.astype(np.float32)
--- a/python/src/spark_rapids_ml/umap.py
+++ b/python/src/spark_rapids_ml/umap.py
@@ -674,6 +674,9 @@ class UMAPModel(_CumlModel, UMAPClass, _UMAPCumlParams):
 
             embedding = self.embedding
             raw_data = self.raw_data
+            if embedding.dtype != np.float32:
+                embedding = embedding.astype(np.float32)
+                raw_data = raw_data.astype(np.float32)
 
             if is_sparse(raw_data):
                 raw_data_cuml = SparseCumlArray(raw_data, convert_format=False)
--- a/python/src/spark_rapids_ml/umap.py
+++ b/python/src/spark_rapids_ml/umap.py
@@ -14,14 +14,19 @@
 # limitations under the License.
 #
 
+import json
+import os
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Dict,
     Generator,
+    Iterable,
+    Iterator,
     List,
     Optional,
+    Sequence,
     Tuple,
     Type,
     Union,
@@ -30,7 +35,9 @@ from typing import (
 import numpy as np
 import pandas as pd
 import pyspark
+from pandas import DataFrame as PandasDataFrame
 from pyspark.ml.param.shared import HasFeaturesCol, HasLabelCol, HasOutputCol
+from pyspark.ml.util import DefaultParamsReader, DefaultParamsWriter, MLReader, MLWriter
 from pyspark.sql import Column, DataFrame
 from pyspark.sql.dataframe import DataFrame
 from pyspark.sql.types import (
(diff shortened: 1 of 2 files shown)


python/src/spark_rapids_ml/umap.py
```diff
@@ -378,6 +409,10 @@ def _fit(self, dataset: DataFrame) -> "UMAPModel":
             dtype=type(raw_data[0][0]).__name__,
         )
 
+        print(
+            type(raw_data[0][0]).__name__,
+        )
```
leewyang, line 414: Is this still needed?  If so, do we want a more descriptive message in the logs?

python/src/spark_rapids_ml/umap.py
```diff
@@ -454,6 +489,111 @@ def _cuml_fit(
...
+            # call the cuml fit function
+            # *note*: cuml_fit_func may delete components of inputs to free
+            # memory.  do not rely on inputs after this call.
+            result = cuml_fit_func(inputs, params)
+            logger.info("Cuml fit complete")
+
+            for row in result:
+                yield row
+
+        output_df = dataset.mapInPandas(_train_udf, schema=self._out_schema())
```
leewyang, line 593: If this is mostly duplicated code, wondering if it can be refactored into the existing API.

Also, is the `fit_multiple_params` API (from @wbo4958) explicitly unsupported then?  If so, maybe we should document this, especially if it's removed for specific reasons.

python/src/spark_rapids_ml/umap.py
```diff
@@ -514,14 +654,14 @@ def __init__(
...
+    def embedding(self) -> np.ndarray:
+        if isinstance(self.embedding_, np.ndarray):
             return self.embedding_
         return self.embedding_.value
 
     @property
-    def raw_data(self) -> List[List[float]]:
-        if isinstance(self.raw_data_, list):
+    def raw_data(self) -> np.ndarray:
+        if isinstance(self.raw_data_, np.ndarray):
```
leewyang, line 664: Pyspark properties generally return python base types.  Wondering if these public APIs should convert to base types here (for external users), while the internal code can still reference the underlying numpy variables, e.g. `self.raw_data_`.  Not sure if others have an opinion...

python/src/spark_rapids_ml/umap.py
```diff
@@ -538,19 +678,22 @@ def _construct_umap() -> CumlT:
 
             from .utils import cudf_to_cuml_array
 
-            embedding_np = np.array(self.embedding, dtype=np.float32)
-            raw_data_np = np.array(self.raw_data, dtype=np.float32)
+            embedding = self.embedding
+            raw_data = self.raw_data
```
leewyang, line 682: If changing the property type per above, would need to modify these accordingly.

python/src/spark_rapids_ml/umap.py
```diff
@@ -538,19 +678,22 @@ def _construct_umap() -> CumlT:
...
+            if embedding.dtype != np.float32:
+                embedding = embedding.astype(np.float32)
+                raw_data = raw_data.astype(np.float32)
```
leewyang, line 685: Probably should log a warning that we're auto-converting the type (but only if it's not too chatty).

python/src/spark_rapids_ml/umap.py
```diff
@@ -596,9 +739,76 @@ def _out_schema(self, input_schema: StructType) -> Union[StructType, str]:
...
     def get_model_attributes(self) -> Optional[Dict[str, Any]]:
-        """Override parent method to bring broadcast variables to driver before JSON serialization."""
-        if not isinstance(self.embedding_, list):
+        """
+        Override parent method to bring broadcast variables to driver before JSON serialization.
+        """
+        if not isinstance(self.embedding_, np.ndarray):
             self._model_attributes["embedding_"] = self.embedding_.value
-        if not isinstance(self.raw_data_, list):
+        if not isinstance(self.raw_data_, np.ndarray):
```
leewyang, line 747: These could be moved to an internal accessor API (if modifying API per above).

leewyang (commented):
Looks nice!   Just some minor comments and some questions for rest of team.

python/src/spark_rapids_ml/umap.py
rishic3: oops, leftover from debugging.

python/src/spark_rapids_ml/umap.py
rishic3: In the case of UMAP, fit_multiple_params isn't used - I basically trimmed out all the bits of call_cuml_fit_func that aren't relevant to UMAP. If we were to make this function more integrated with the existing API, definitely could add back in those options.

As for refactoring this into the existing API, don't think there's a clean way without overriding or creating a new call_fit_func in core due to the RDD return signature and the barrier stuff. If we're interested in using dataframes for future algos that don't require NCCL during fit, we could have a second call_cuml_fit_func within core like the one in this PR for those use cases which future algos (and this algo) could inherit from. Not sure if this is preferred,

python/src/spark_rapids_ml/umap.py
rishic3: At the moment I'm not supporting user-control over the "convert_dtype" param from cuml (determines whether the internal computations are float64); currently just defaulting to float32. I figured we could keep it like that for now and add float64 support in a future pr (and like Erik mentioned, maybe include a default conversion to float32 much earlier in the computation if desired).

python/src/spark_rapids_ml/umap.py
rishic3: In that case, probably would be good return it list-converted without changing the internal references.

python/src/spark_rapids_ml/umap.py
rishic3: I think this is needed in case the user never accesses the attributes before saving - attributes will need to be retrieved and stored from the broadcast.

wbo4958:
@rishic3 Could you run 

`rdd + no-repartition + collect (no overrides)` by adding 

`--conf spark.sql.files.minPartitionNum=$gpu_workers  --conf spark.sql.files.maxPartitionBytes=50000000000`

leewyang:
build

rishic3:
> @rishic3 Could you run
> 
> `rdd + no-repartition + collect (no overrides)` by adding
> 
> `--conf spark.sql.files.minPartitionNum=$gpu_workers --conf spark.sql.files.maxPartitionBytes=50000000000`

Both tests were run with these settings. I used the benchmark spark config

leewyang (approved):

PR #361, opened 2023-08-10
add larger dataset demo to umap notebook

rishic3:


leewyang:
Looks nice!  A couple questions:
- it looks like the new visualization only has one color, shouldn't it theoretically show three different colored blobs? (i.e. the cmap looks like it's still 10 classes, but probably should be 3).
- is it still possible to run/time/visualize the local CPU UMAP on the larger dataset?

leewyang (approved):

PR #372, opened 2023-08-15
fix conda install commands

eordentlich:
avoid cuda-toolkit and use [specified channel order](https://docs.rapids.ai/install#selector) which is apparently critical for correct cuda12/cuda11 dependencies.
23.08 will move to cuda-version from cudatoolkit so switched to that.

### DIFF
--- a/docker/Dockerfile.python
+++ b/docker/Dockerfile.python
@@ -38,7 +38,7 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py38_4.10.3-Linu
 
 # install cuML
 
-RUN conda install -y -c rapidsai -c nvidia -c conda-forge python=3.9 cuda-toolkit=11.5 cuml=$CUML_VERSION \
+RUN conda install -y -c rapidsai -c conda-forge -c nvidia python=3.9 cudatoolkit=11.5 cuml=$CUML_VERSION \
     && conda clean --all -f -y
 
 # install python dependencies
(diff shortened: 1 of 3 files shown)


eordentlich:
build

leewyang (commented):
Any reason we aren't using `cuda-version` in the `Dockerfile.python` and `README.md` also?

docker/Dockerfile.python
```diff
@@ -38,7 +38,7 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py38_4.10.3-Linu
 
 # install cuML
 
-RUN conda install -y -c rapidsai -c nvidia -c conda-forge python=3.9 cuda-toolkit=11.5 cuml=$CUML_VERSION \
+RUN conda install -y -c rapidsai -c conda-forge -c nvidia python=3.9 cudatoolkit=11.5 cuml=$CUML_VERSION \
```
lijinf2, line 41: Does "cudatoolkit" work the same as "cuda-toolkit"?

eordentlich: I don't think so.  cudatoolkit is what the rapids selector has shown.

eordentlich:
> Any reason we aren't using `cuda-version` in the `Dockerfile.python` and `README.md` also?

I'm not sure if cuda-version works for 23.06 which those are still on.   Should be pretty soon to transition to 23.08 though.

lijinf2 (approved):
LGTM

eordentlich:
> Does the cudatoolkit include 'cuda-python, cupy' inside ?

No.  Those are separate python packages.

PR #377, opened 2023-08-16
Support stage-level scheduling for transform

wbo4958:
Users can configure how many transform tasks are running at the same time by introducing the configuration "spark.rapids.ml.concurrentPredictTasks", the default is 2.

### DIFF
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -56,6 +56,7 @@ from pyspark.ml.util import (
     MLWritable,
     MLWriter,
 )
+from pyspark.resource import ResourceProfileBuilder, TaskResourceRequests
 from pyspark.sql import Column, DataFrame
 from pyspark.sql.functions import col, struct
 from pyspark.sql.pandas.functions import pandas_udf
@@ -224,6 +225,7 @@ class _CumlModelReader(MLReader):
 class _CumlCommon(MLWritable, MLReadable):
     def __init__(self) -> None:
         super().__init__()
+        self.logger = get_logger(self.__class__)
 
     @staticmethod
     def set_gpu_device(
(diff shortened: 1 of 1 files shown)


python/src/spark_rapids_ml/core.py
```diff
@@ -1041,7 +1061,69 @@ def _transform_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
+
+        executor_cores = int(ss.sparkContext.getConf().get("spark.executor.cores"))  # type: ignore
+
+        if (
+            executor_cores // 2 < concurrent_transform_tasks < executor_cores
+            or concurrent_transform_tasks > executor_cores
+        ):
+            original_concurrent_transform_tasks = concurrent_transform_tasks
+            concurrent_transform_tasks = (
+                executor_cores // 2
```
eordentlich, line 1092: Is it possible to achieve exactly concurrent_transform_tasks by setting executor_cores=concurrent_transform_tasks and task_cores=1 just for this stage?   Or is executor_cores not configurable per stage?

python/src/spark_rapids_ml/core.py
```diff
@@ -1041,7 +1061,69 @@ def _transform_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
+                if concurrent_transform_tasks < executor_cores
+                else executor_cores
+            )
+            self.logger.warning(
+                "The spark.rapids.ml.concurrentPredictTasks should be in "
+                f"[1, {executor_cores // 2}] or {executor_cores}, "
+                f"we found it's {original_concurrent_transform_tasks}. We're "
+                f"going to use {concurrent_transform_tasks}"
+            )
```
leewyang, line 1101: Should we just raise an error w/ a suggested value instead of modifying a user-supplied conf?

python/src/spark_rapids_ml/core.py
```diff
@@ -1041,7 +1061,69 @@ def _transform_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
+
+        task_cores = executor_cores // concurrent_transform_tasks
+
+        default_task_cpus = int(ss.sparkContext.getConf().get("spark.task.cpus"))  # type: ignore
+        if task_cores == default_task_cpus:
+            self.logger.warning(
+                "The Spark configurations have been set up to ensure that "
+                f"{concurrent_transform_tasks} transform tasks will be concurrently executed."
+            )
```
leewyang, line 1110: Should this be `info` instead of `warning`?

python/src/spark_rapids_ml/core.py
wbo4958: Actually, we can't require executor resources which will cause allocating new executors. But in our case, we don't want to require more executores.

python/src/spark_rapids_ml/core.py
wbo4958: Good question. My original thinking is "the train + transform may be a typical ML case, if we raise an error in transform, the user might waste time re-train again.

python/src/spark_rapids_ml/core.py
wbo4958: Good catch.

python/src/spark_rapids_ml/core.py
```diff
@@ -1041,7 +1061,69 @@ def _transform_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
+        self.logger.info(
+            f"Transform tasks require the resource(cores={task_cores}, gpu=1.0) and there will be "
+            f"{concurrent_transform_tasks} transform tasks running at the same time."
+        )
+
+        # TODO Do we need to coalesce partitions to num_workers*concurrent_transform_tasks
+        # num_workers = self._infer_num_workers()
+        # df = df.coalesce(num_workers * concurrent_transform_tasks)
+
+        df = df.rdd.withResources(rp).toDF()
```
eordentlich, line 1125: Is there a perf impact of df -> rdd -> df  re. Spark RAPIDS plugin?

leewyang: Maybe should log a warning (w/ recommended settings) if there is an expected perf impact?

python/src/spark_rapids_ml/core.py
```diff
@@ -278,6 +280,66 @@ def initialize_cuml_logging(verbose: Optional[Union[bool, int]]) -> None:
...
+            )
+            return True
+
+        task_gpu_amount = sc.getConf().get("spark.task.resource.gpu.amount")
+
+        if task_gpu_amount is None:
+            # if spark.task.resource.gpu.amount is not set, the default concurrent tasks
+            # with gpu requirement will be 1, which means 2 training tasks will never
+            # be scheduled into the same executor.
+            return True
```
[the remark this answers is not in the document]
wbo4958: Yeah, good suggestion, added the warning.

python/src/spark_rapids_ml/core.py
wbo4958: Cool, Added the warning in the log.

wbo4958:
BTW, Could you give me some suggestions on the naming of "spark.rapids.ml.concurrentPredictTasks", I don't think it's a good name, but I can't think of a good one.

wbo4958:
close it temporary

PR #375, opened 2023-08-15
[Logistic Regression] Support benchmarking in run_benchmark.sh

lijinf2:

PR #381, opened 2023-08-17
[Logistic Regression] set hasSummary to false to be compatible with spark

lijinf2:
A followup PR of https://github.com/NVIDIA/spark-rapids-ml/pull/375


lijinf2:
build

PR #382, opened 2023-08-17
add umap to benchmarking

rishic3:
also added some comments about dtype support and visualization differences to notebook, and minor max_records_per_batch change in umap.

### DIFF
--- a/python/benchmark/benchmark/bench_umap.py
+++ b/python/benchmark/benchmark/bench_umap.py
@@ -1,0 +1,170 @@
+#
+# Copyright (c) 2023, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import time
+from typing import Any, Dict, List, Optional, Union
+
+from pyspark.ml.feature import VectorAssembler
+from pyspark.ml.functions import array_to_vector, vector_to_array
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql.functions import col, sum
+
+from benchmark.base import BenchmarkBase
+from benchmark.utils import with_benchmark
+
+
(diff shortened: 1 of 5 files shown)


python/src/spark_rapids_ml/umap.py
```diff
@@ -343,7 +343,7 @@ def __init__(self, sample_fraction: float = 1.0, **kwargs: Any) -> None:
         super().__init__()
         self.set_params(**kwargs)
         self.sample_fraction = sample_fraction
-        self.maxRecordsPerBatch = 10000
+        self.max_records_per_batch = 10000
```
lijinf2, line 346: Can we set this in the same way as line 369, i.e. through "spark.sql.execution.arrow.maxRecordsPerBatch"?

lijinf2, line 346: Can we set this in the same way as line 370, i.e. through "spark.sql.execution.arrow.maxRecordsPerBatch"?

python/benchmark/benchmark/bench_umap.py
```diff
@@ -0,0 +1,170 @@
...
+            print(f"cpu total took: {total_time} sec")
+
+        report_dict = {
+            "fit": fit_time,
+            "transform": transform_time,
+            "total_time": total_time,
+            "num_gpus": num_gpus,
+            "num_cpus": num_cpus,
+            "no_cache": no_cache,
+            "train_path": self.args.train_path,
```
lijinf2, line 164: Is it possible to add one or two metrics to show the quality of umap and pca?

python/run_benchmark.sh
```diff
@@ -390,3 +391,61 @@ if [[ "${MODE}" =~ "logistic_regression" ]] || [[ "${MODE}" == "all" ]]; then
...
+        --regParam 0.00001 \
+        --num_gpus $num_gpus \
+        --num_cpus $num_cpus \
+        --num_runs $num_runs \
+        --train_path "${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet" \
+        --transform_path "${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet" \
+        --report_path "report_logistic_regression_${cluster_type}.csv" \
+        $common_confs $spark_rapids_confs \
+        ${EXTRA_ARGS}
+fi
```
leewyang, line 425: Not sure why LogisticRegression benchmark shows as a change, does this PR need to be re-based now that the other PR is merged?

python/src/spark_rapids_ml/umap.py
leewyang: Basically, this line looks redundant w/ L369-373.  Not sure if that's intentional.

PR #386, opened 2023-08-17
added umap to benchmarking

rishic3:
#### benchmarking
- compares umap with spark pca
- returns trustworthiness score for each
- using blobs dataset. isn't quite the ideal use case for either algo, but probably the best choice out of gen_data options
   * umap usually designed for more complex datasets
   * pca focused on overall variance, heavily influenced by cluster-separation

#### notebook
- minor comments about dtype and cpu/gpu visualization differences
- removed personal user info from pip output

#### umap
- minor change to set `max_records_per_batch` upon init


leewyang (approved):

rishic3:
build

PR #388, opened 2023-08-18
add logistic regression to cloud bms

eordentlich:


eordentlich:
changes do not impact pre-merge ci so can skip

python/run_benchmark.sh
```diff
@@ -362,10 +362,10 @@ fi
 # Logistic Regression Classification
 if [[ "${MODE}" =~ "logistic_regression" ]] || [[ "${MODE}" == "all" ]]; then
     if [[ ! -d ${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet ]]; then
+    # same data set as random forest classifier
```
leewyang, line 365: assuming this is intentional?

leewyang (approved):

python/run_benchmark.sh
lijinf2: I think so. LR benchmark can reuse the classification dataset of random forest classifier.

lijinf2 (approved):
Looks good!

PR #404, opened 2023-08-24
Umap notebook fixes

rishic3:
Fixed an issue in the notebook where same input was being fed for fit and transform on the CPU UMAP side, resulting in massive runtime underestimation of CPU UMAP during transform by spitting out the trained embeddings (see cuml's [hash_input param](https://docs.rapids.ai/api/cuml/stable/api/#umap)). 

After fixes, Spark-RAPIDS-ML UMAP fit is ~**1.7x** faster during fit, and ~**92x** faster during transform compared to CPU on the larger dataset.


rishic3:
build

leewyang (approved):

eordentlich:
build

PR #405, opened 2023-08-25
preserve input column format and some other fixes

eordentlich:
- preserves input column format in transform
- more float64 to float32 conversion for other formats in logistic regression
- prediction column returned as float64
- remove float64 casting in benchmark scripts given previous change.


eordentlich:
build

eordentlich:
build

leewyang (approved):

PR #353, opened 2023-08-07
Upgrade cuml from 23.06 to 23.08 in pytests and ci

lijinf2:
Fixed an issue in LinearRegression that adds the copy_X init argument to match cuml 23.08 LinearRegression init argument list.

PR #407, opened 2023-08-25
add copy_X=False to transform linear regression

eordentlich:
This eliminates the below warning during transforms.  A [previous PR](https://github.com/NVIDIA/spark-rapids-ml/pull/353/files#diff-d734f3d25daffabbee57552d24e4d2280576ce5fc7f13c6dd7e41c443a5efb6cR554) addressed fit.
```UserWarning: Starting from version 23.08, the new 'copy_X' parameter defaults to 'True', ensuring a copy of X is created after passing it to fit(), preventing any changes to the input, but with increased memory usage. This represents a change in behavior from previous versions. With `copy_X=False` a copy might still be created if necessary. Explicitly set 'copy_X' to either True or False to suppress this warning.```

### DIFF
--- a/python/src/spark_rapids_ml/regression.py
+++ b/python/src/spark_rapids_ml/regression.py
@@ -712,7 +712,7 @@ class LinearRegressionModel(
             intercepts = intercept_ if isinstance(intercept_, list) else [intercept_]
 
             for i in range(len(coefs)):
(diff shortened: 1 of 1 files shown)


eordentlich:
build

leewyang (approved):

PR #417, opened 2023-08-30
RF classifier: map rawPredictionCol to empty and show the warning

wbo4958:
spark-rapids-ml doesn't support rawPredictionCol, this PR just remove it.

### DIFF
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -124,8 +124,16 @@ class _RFClassifierParams(
         return self._set(rawPredictionCol=value)
 
 
+class _RandomForestClassifierClass(_RandomForestClass):
+    @classmethod
+    def _param_mapping(cls) -> Dict[str, Optional[str]]:
+        mapping = super()._param_mapping()
+        mapping["rawPredictionCol"] = ""
+        return mapping
+
+
 class RandomForestClassifier(
-    _RandomForestClass,
+    _RandomForestClassifierClass,
     _RandomForestEstimator,
     _RandomForestCumlParams,
     _RFClassifierParams,
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -172,8 +172,6 @@ class RandomForestClassifier(
         The prediction column name.
     probabilityCol
         The column name for predicted class conditional probabilities.
-    rawPredictionCol:
-        The raw prediction column name.
     maxDepth:
         Maximum tree depth. Must be greater than 0.
     maxBins:
@@ -287,7 +285,6 @@ class RandomForestClassifier(
         labelCol: str = "label",
         predictionCol: str = "prediction",
         probabilityCol: str = "probability",
-        rawPredictionCol: str = "rawPrediction",
         maxDepth: int = 5,
         maxBins: int = 32,
         minInstancesPerNode: int = 1,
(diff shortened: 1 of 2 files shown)


wbo4958:
build

eordentlich (commented):
Should constructor arg for this be removed as well?

wbo4958:
> Should constructor arg for this be removed as well?

Yeah, you're right. will fix it.

wbo4958:
build

wbo4958:
build

wbo4958:
build

wbo4958:
we have a test case to test the spark compatibility, so we can't just simply remove rawPredictionCol though cuml doesn't support it. Instead, mapping rawPredictionCol to empty and get the unsupported warning.

eordentlich (commented):
What would getter do?

On second thought, looks like probability column is simple invertible function of raw prediction column per this SO post: https://stackoverflow.com/questions/37903288/what-do-columns-rawprediction-and-probability-of-dataframe-mean-in-spark-mll
Maybe we can just invert the function and add the column using cupy or numpy or even DF operations?

wbo4958:
> What would getter do?
> 
> On second thought, looks like probability column is simple invertible function of raw prediction column per this SO post: https://stackoverflow.com/questions/37903288/what-do-columns-rawprediction-and-probability-of-dataframe-mean-in-spark-mll Maybe we can just invert the function and add the column using cupy or numpy or even DF operations?

Yeah, good idea. But could we delay this to the next release? we're approaching the deadline.

eordentlich:
Ok.   But I think if you retain HasRawPredictionCol, the getter will exist and return the default value.   Do we want that?

wbo4958:
> Ok. But I think if you retain HasRawPredictionCol, the getter will exist and return the default value. Do we want that?

Yes, that's expected. we've a test case to cover this https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.08/python/tests/test_random_forest.py#L575-L576

eordentlich:
Can you explain the goal of this PR in that case (the setter and getter still work, but constructor arg is ignored?) ?   

Also this and LogisticRegression classifier should be consistent on this (it also doesn't yet support computing rawPredictionCol).

wbo4958:
> the setter and getter still work, but constructor arg is ignored?

No, the constructor will still set the rawPredictionCol value, but it will not be used in cuml side, so we map it to an empty string.

I thought it's the compatibility case. Consider the case, that the user would like to try spark-rapids-ml by just replacing the pyspark imports, and what if the user's case just has the rawPredictionCol in the constructor? In that case, spark-rapids-ml will throw an exception if we remove rawPredictionCol, instead, I think it would be better if we just warn the user that rawPredictionCol is not being used.

eordentlich:
Ok.  Unlikely a user will specify a column for this and then not actually use it.   But this at least should remove from our api docs for now.   Can you extend this change in this PR also to logistic regression which has the same issue:
https://github.com/NVIDIA/spark-rapids-ml/pull/417/files#diff-5bffca677cda02544897eeac447779db91115ec59138bc10c6115e56ed123692R131

wbo4958:
@eordentlich could you help to review again. BTW, I found the weightCol and leafCol are mapped to None. Do I need to map the rawPredictionCol to None?

wbo4958:
build

eordentlich (approved):
👍

PR #419, opened 2023-08-30
Enable stage-level scheduling only when it's on the standalone cluster

wbo4958:
stage-level scheduling with dynamic allocation disabled only works on the standalone and local-cluster mode according to https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/resource/ResourceProfileManager.scala#L70


wbo4958:
build

eordentlich (approved):
LGTM

PR #422, opened 2023-08-30
include logistic regression benchmark

eordentlich:
same setting as other (2 node databricks, 12GB dataset).

and add pointers for other CSPs.


eordentlich:
NO CI needed.

PR #435, opened 2023-09-06
raise exception in python worker receiving no data

eordentlich:


eordentlich:
build

leewyang (approved):
LGTM.  Just curious, could we return an empty result/dataframe instead (i.e. skipping the `cuml_fit_func`)?

eordentlich:
All workers would need to coordinate to resize nccl communicator world to only those workers with data.   
Seems overkill for now for tiny data examples involving multiple workers.

PR #438, opened 2023-09-07
RF test: relax the validation restriction for cpu and gpu predictions

wbo4958:


wbo4958:
build

python/tests/test_random_forest.py
```diff
@@ -359,11 +359,15 @@ def test_random_forest_classifier(
         result = spark_rf_model.transform(test_df).collect()
         pred_result = [row.prediction for row in result]
 
+        # no need to compare all feature types.
         if feature_type == feature_types.vector:
-            # no need to compare all feature type.
             spark_cpu_result = spark_rf_model.cpu().transform(test_df).collect()
             spark_cpu_pred_result = [row.prediction for row in spark_cpu_result]
-            assert array_equal(spark_cpu_pred_result, pred_result)
+            # Since the precision issue, we can ensure all the predictions are same.
```
leewyang, line 366: "Since cuML uses FP32, but Spark uses FP64, we cannot ensure that all predictions will be the same."?

leewyang: Otherwise LGTM

leewyang (approved):

python/tests/test_random_forest.py
wbo4958: I will update the comment in the following PR. For now, let's just merge it.

PR #440, opened 2023-09-07
Validate parameters by converting them to spark java companion objects

wbo4958:


wbo4958:
build

leewyang (approved):
LTGM.  Wondering if we should explicitly delete the spark LR object after the validations, just in case it's holding on to any resources.

wbo4958:
> LTGM. Wondering if we should explicitly delete the spark LR object after the validations, just in case it's holding on to any resources.

Good idea.

wbo4958:
I realized this PR can just validate the parameters in the `__ int __`, but it can't validate the parameter when the user calls the set function. I will rework it a little bit.

wbo4958:
build

wbo4958:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -400,6 +401,57 @@ def _require_nccl_ucx(self) -> Tuple[bool, bool]:
...
+            pyspark_est = PCA()
+        elif "KMeans" in cls_name:
+            from pyspark.ml.clustering import KMeans
+
+            pyspark_est = KMeans()
+        if pyspark_est is not None:
+            # Both pyspark and cuml may have a parameter with the same name,
+            # but cuml might have additional optional values that can be set.
+            # If we transfer these cuml-specific values to the Spark JVM,
+            # it would result in an exception.
```
eordentlich, line 438: Looks like we currently allow setting cuml param directly in which case we also set the corresponding Spark param, but to the cuml-specific value:  https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.08/python/src/spark_rapids_ml/params.py#L267
(see the TODO).   Not sure if this is actually used though.

python/src/spark_rapids_ml/core.py
```diff
@@ -400,6 +401,57 @@ def _require_nccl_ucx(self) -> Tuple[bool, bool]:
...
+        elif "PCA" in cls_name:
+            from pyspark.ml.feature import PCA
+
+            pyspark_est = PCA()
+        elif "KMeans" in cls_name:
+            from pyspark.ml.clustering import KMeans
+
+            pyspark_est = KMeans()
+        if pyspark_est is not None:
+            # Both pyspark and cuml may have a parameter with the same name,
```
eordentlich, line 435: In this case, wouldn't the value stored in the Spark param be the unmapped value and not the cuml-specific value?   So seems it should be ok.   It's the other case that might be a problem (setting a cuml param value directly) - see other comment.

python/src/spark_rapids_ml/core.py
```diff
@@ -400,6 +401,57 @@ def _require_nccl_ucx(self) -> Tuple[bool, bool]:
...
         return (True, False)
 
+    def _validate_parameters(self) -> None:
+        cls_name = self.__class__.__name__
+
+        pyspark_est: Optional[JavaParams] = None
+
+        # Is there a better way for the dynamic imports?
+        if "LogisticRegression" in cls_name:
+            from pyspark.ml.classification import LogisticRegression
```
leewyang, line 411: Think we can add a method to `_CumlClass` to hold the pyspark/cpu class type, s.t. each implementation specifies its own pyspark/cpu equivalent.

eordentlich:
Seems best to postpone this patch to 23.10 to allow more time to improve/verify.

eordentlich:
@wbo4958 Can you redirect this PR to branch-23.10 ?

python/src/spark_rapids_ml/core.py
wbo4958: Yeah. Good idea. Maybe just let subclasses validate themself?

python/src/spark_rapids_ml/core.py
eordentlich: This is a good way to handle this issue.

PR #443, opened 2023-09-08
add some missing updates to dataproc README [skip ci]

eordentlich:
we no longer have our own version of spark-rapids.sh


leewyang (approved):
Can likely use hosted version directly: gs://goog-dataproc-initialization-actions-us-central1/rapids/rapids.sh

eordentlich:
I've found the gcp stored versions to be quite out of date.

PR #455, opened 2023-09-29
databricks updates

eordentlich:
- runtime from 11.3 to 12.2
- cluster init script in workspace
- latest version of databricks cli (old one didn't support workspace imports for non-python scripts)
- use 2.1 api for clusters, runs, etc.  as latest cli only supports 2.1


eordentlich:
No CI needed.

leewyang (approved):

PR #457, opened 2023-10-02
revise knn test to better compensate for numerical variations between kernels

eordentlich:
Was able to reproduce numerical discrepancies for self distance 
via direct calls to raft distance computation methods.

Revised test to use smaller norm vectors and also do the comparison in the 'squared distance' domain so that the numerical errors are within our default tolerances (the latter could also have been scaled up).

Interestingly RAFT does some clamping to 0 of small distances, but it is not scale independent:
https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/knn_brute_force.cuh#L202

The average unit norm change should put the vectors in that expected scale.

### DIFF
--- a/python/tests/test_nearest_neighbors.py
+++ b/python/tests/test_nearest_neighbors.py
@@ -326,6 +326,11 @@ def test_nearest_neighbors(
         random_state=0,
     )  # make_blobs creates a random dataset of isotropic gaussian blobs.
 
+    # set average norm to be 1 to allow comparisons with default error thresholds
+    # below
+    ave_norm_X = np.average(np.linalg.norm(X, ord=2, axis=1))
+    X = X/ave_norm_X
+
     # obtain cuml results
     from cuml import NearestNeighbors as cuNN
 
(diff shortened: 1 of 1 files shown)


eordentlich:
build

eordentlich:
build

eordentlich:
build

python/tests/test_nearest_neighbors.py
```diff
@@ -364,10 +369,18 @@ def test_nearest_neighbors(
             item_df_withid.select(alias.row_number).toPandas()[alias.row_number]
         )
 
-        # test kneighbors: compare distances
+        # test kneighbors: compare squared distances
+        # note that single node and multi node may run slightly different kernels resulting
+        # in different distances.  This is especially an issue for self distances which don't come out
+        # to be 0 necessarily due to expanded form of calculation (|x-y|^2 = |x| + |y| - 2 <x,y>).
```
lijinf2, line 375: Not quite understand but feel |x-y|^2 = `|x|^2` + `|y|^2` - 2 <x,y>. Or it should be `|x|`, `|y|` instead of `|x|^2` and `|y|^2`. 

Seems like `|x|` and `|y|` are used to approximate their squared value.

eordentlich:
build

python/tests/test_linear_model.py
lijinf2, line 1: This file gets removed by the latest commit. Seems it is by accident. Otherwise the PR looks good!

eordentlich: Oops.   Good catch.   Was debugging linear regression in same branch and didn't fully restore the branch.   Fixed

eordentlich:
build

lijinf2 (approved):

PR #460, opened 2023-10-04
[Logistic Regression] Support multiple classes

lijinf2:
Support multiple classes in fit, transform, and benchmarking.

### DIFF
--- a/python/benchmark/benchmark/bench_logistic_regression.py
+++ b/python/benchmark/benchmark/bench_logistic_regression.py
@@ -86,6 +86,7 @@ class BenchmarkLogisticRegression(BenchmarkBase):
             print(f"total iterations: {model.summary.totalIterations}")
             print(f"objective history: {model.summary.objectiveHistory}")
         else:
+            print(f"total iterations: {model.num_iters}")
             print("model does not have hasSummary attribute")
 
         eval_df = train_df if transform_df is None else transform_df
@@ -163,7 +164,6 @@ class BenchmarkLogisticRegression(BenchmarkBase):
                 .setLabelCol(label_name)
             )
 
-            # default to accuracy metric
             metric_value = evaluator_test.evaluate(eval_df_with_preds)
 
             print(
--- a/python/benchmark/benchmark/bench_logistic_regression.py
+++ b/python/benchmark/benchmark/bench_logistic_regression.py
@@ -132,7 +132,11 @@ class BenchmarkLogisticRegression(BenchmarkBase):
 
         # TODO: add l1 regularization penalty term to full objective for when we support it
         log_loss = evaluator_train.evaluate(train_df_with_preds)
-        coefficients = model.coefficientMatrix.toArray()
+        coefficients = (
+            np.array(model.coefficients)
+            if model.numClasses == 2
+            else model.coefficientMatrix.toArray()
+        )
         coefs_l1 = np.sum(np.abs(coefficients))
         coefs_l2 = np.sum(coefficients**2)
 
--- a/python/benchmark/benchmark/bench_logistic_regression.py
+++ b/python/benchmark/benchmark/bench_logistic_regression.py
@@ -109,44 +109,9 @@ class BenchmarkLogisticRegression(BenchmarkBase):
             MulticlassClassificationEvaluator,
         )
 
-        # TODO: support multiple classes
-        # binary classification
-        evaluator_train: Union[
-            BinaryClassificationEvaluator, MulticlassClassificationEvaluator
-        ] = (
-            MulticlassClassificationEvaluator()
-            .setMetricName("logLoss")  # type:ignore
-            .setPredictionCol(prediction_col)
-            .setProbabilityCol(probability_col)
-            .setLabelCol(label_name)
-        )
-
-        evaluator_test: Union[
-            BinaryClassificationEvaluator, MulticlassClassificationEvaluator
-        ] = (
-            BinaryClassificationEvaluator()
-            .setRawPredictionCol(probability_col)
-            .setLabelCol(label_name)
-        )
-
-        log_loss = evaluator_train.evaluate(train_df_with_preds)
-        coefficients = np.array(model.coefficients)
-        coefs_l1 = np.sum(np.abs(coefficients))
-        coefs_l2 = np.sum(coefficients**2)
-
-        # TODO: add l1 regularization penalty term to full objective for when we support it
-        train_full_objective = log_loss + 0.5 * lr.getRegParam() * coefs_l2
-
-        eval_auc = evaluator_test.evaluate(eval_df_with_preds)
-
-        print(f"{benchmark_string} train_full_objective: {train_full_objective}")
-        print(f"{benchmark_string} eval_auc: {eval_auc}")
-
         results = {
             "fit_time": fit_time,
             "transform_time": transform_time,
-            "train_full_objective": train_full_objective,
-            "eval_auc": eval_auc,
             "num_gpus": self.args.num_gpus,
             "num_cpus": self.args.num_cpus,
             "train_path": self.args.train_path,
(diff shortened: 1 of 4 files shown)


lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -984,44 +990,85 @@ def coefficients(self) -> Vector:
...
-            ).array
-            lr.coef_ = input_to_cuml_array(
-                np.array(coef_, order="C").astype(dtype)
-            ).array
-            # TBD: infer class indices from data for > 2 classes
-            # needed for predict_proba
+
+            gpu_intercept_ = cp.array(intercept_, order="C", dtype=dtype)
+            gpu_coef_ = cp.array(coef_, order="F", dtype=dtype).T
+            gpu_stacked = GlobalSettings().xpy.vstack([gpu_coef_, gpu_intercept_])
```
eordentlich, line 1058: Does cupy not have vstack?

python/src/spark_rapids_ml/classification.py
```diff
@@ -984,44 +990,85 @@ def coefficients(self) -> Vector:
...
+            lr.solver_model._coef_ = input_to_cuml_array(gpu_stacked, order="C").array
+
+            lr.penalty_normalized = False
+            lr.lbfgs_memory = 10
```
eordentlich, line 1062: Are these needed for transform?

python/src/spark_rapids_ml/classification.py
```diff
@@ -984,44 +990,85 @@ def coefficients(self) -> Vector:
...
+
             lr.classes_ = input_to_cuml_array(
-                np.array([0, 1], order="F").astype(dtype)
+                np.array(classes_, order="F").astype(dtype)
             ).array
+            lr._num_classes = len(lr.classes_)
+
+            lr.loss = "sigmoid" if lr._num_classes <= 2 else "softmax"
+            lr.solver_model.qnparams = lr.create_qnparams()
+
```
eordentlich, line 1071: Needed for transform?

eordentlich (commented):
Looks good.   A few minor comment.s

python/src/spark_rapids_ml/classification.py
```diff
@@ -54,7 +54,7 @@
     _LogisticRegressionParams,
     _RandomForestClassifierParams,
 )
-from pyspark.ml.linalg import DenseMatrix, Vector, Vectors
+from pyspark.ml.linalg import DenseMatrix, Matrices, Matrix, Vector, Vectors
```
leewyang, line 57: Looks like `Matrices` is unused?

python/src/spark_rapids_ml/classification.py
```diff
@@ -984,44 +990,85 @@ def coefficients(self) -> Vector:
...
+        classes_ = self.classes_
         n_cols = self.n_cols
         dtype = self.dtype
 
         def _construct_lr() -> CumlT:
+            import cupy as cp
             import numpy as np
-            from cuml.internals.input_utils import input_to_cuml_array
+            from cuml.internals.global_settings import GlobalSettings
+            from cuml.internals.input_utils import CumlArray, input_to_cuml_array
```
leewyang, line 1049: Are there any equivalent APIs that aren't in the `cuml.internals` namespace?

leewyang:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -984,44 +985,85 @@ def coefficients(self) -> Vector:
         """
         Model coefficients.
         """
-        assert len(self.coef_) == 1, "multi classes not supported yet"
-        return Vectors.dense(cast(list, self.coef_[0]))
+        if len(self.coef_) == 1:
+            return Vectors.dense(cast(list, self.coef_[0]))
+        else:
+            raise Exception(
```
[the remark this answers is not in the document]
lijinf2: Yes, I copied the Exception Message from Spark CPU output. 

Added the match checking to compatible tests. Will be in next PR.

python/benchmark/benchmark/bench_logistic_regression.py
```diff
@@ -121,39 +132,43 @@ def run_once(
...
-        ] = (
-            BinaryClassificationEvaluator()
-            .setRawPredictionCol(probability_col)
-            .setLabelCol(label_name)
-        )
-
+        # TODO: add l1 regularization penalty term to full objective for when we support it
         log_loss = evaluator_train.evaluate(train_df_with_preds)
-        coefficients = np.array(model.coefficients)
+        coefficients = model.coefficientMatrix.toArray()
```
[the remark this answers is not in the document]
lijinf2: binomial returns a 2D matrix, e.g. [[0.xxx, 0.xxx]]. The regularization computation is not explained in the Spark doc for the binomial case, so I followed the doc of multinomial case: https://spark.apache.org/docs/latest/ml-classification-regression.html#multinomial-logistic-regression

Do you mean we should remove the regularization computation for the binomial case?

python/run_benchmark.sh
```diff
@@ -362,17 +363,24 @@ fi
 
 # Logistic Regression Classification
 if [[ "${MODE}" =~ "logistic_regression" ]] || [[ "${MODE}" == "all" ]]; then
-    if [[ ! -d ${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet ]]; then
+    data_path=${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet
+
+    if [[ $num_classes != 2 ]]; then
```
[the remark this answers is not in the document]
lijinf2: Revised to set logistic regression separated and default number of classes to 10.

python/tests/test_logistic_regression.py
```diff
@@ -469,3 +479,62 @@ def test_lr_fit_multiple_in_single_pass(
...
+def test_multiclass(
+    fit_intercept: bool,
+    feature_type: str,
+    data_shape: Tuple[int, int],
+    data_type: np.dtype,
+    max_record_batch: int,
+    n_classes: int,
+    gpu_number: int,
+) -> None:
+    tolerance = 0.01
```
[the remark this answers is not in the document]
lijinf2: The number of model parameters increases as more classes. 
Tolerance 0.005 works well for 4 classes. But when 8 classes, the 0.005 leads to several value mismatches when assert array equal of Spark GPU coefficients and Cuml SG coefficients. So the tolerance was relaxed to 0.01.

python/src/spark_rapids_ml/classification.py
lijinf2: Revised to use cupy. Just found it does have vstack.

python/src/spark_rapids_ml/classification.py
lijinf2: Seems not needed. Just to be consistent with fit

python/src/spark_rapids_ml/classification.py
lijinf2: Yes, it is required. cuml relies on the qnparams to decide whether 2-classes or multiple classes. 

Meanwhile, lr.loss has to be set before create_qnparams, because lr.loss is used to create a correct qnparams.

python/benchmark/benchmark/bench_logistic_regression.py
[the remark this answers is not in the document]
eordentlich: Don't remove.  The regularization for the binomial case should involve only coefficients going into the logistic function.   So use the coefficients field and not coefficientMatrix of the model object.

python/run_benchmark.sh
```diff
@@ -362,17 +363,20 @@ fi
 
 # Logistic Regression Classification
 if [[ "${MODE}" =~ "logistic_regression" ]] || [[ "${MODE}" == "all" ]]; then
-    if [[ ! -d ${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32.parquet ]]; then
-    # same data set as random forest classifier
+
+    data_path=${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32_ncls${num_classes}.parquet
```
eordentlich, line 367: So with this we have just num_classes > 2?   We should retain a run with num_classes=2.

python/src/spark_rapids_ml/classification.py
eordentlich: Then better to remove here to avoid confusion.

python/tests/test_logistic_regression.py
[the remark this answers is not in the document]
eordentlich: I don't see why the number of classes should affect the accuracy.   Is it within 1 percent?   Does increasing regularization reduce error?

python/benchmark/benchmark/bench_logistic_regression.py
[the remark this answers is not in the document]
lijinf2: Revised to use coefficients field for the binomial case.

python/tests/test_logistic_regression.py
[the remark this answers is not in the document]
lijinf2: Yes, increasing regularization reduces the error to below 0.005 tolerance. Good idea.

python/src/spark_rapids_ml/classification.py
lijinf2: Removed.

python/run_benchmark.sh
eordentlich: Can we retain a run with num_classes=2 in hte benchmark script (that's an important special case).

lijinf2: Yeah, just revised the PR to run both 2 and 10 classes for LR default.

lijinf2:
build

python/run_benchmark.sh
```diff
@@ -364,36 +364,39 @@ fi
...
-        --num_cpus $num_cpus \
-        --num_runs $num_runs \
-        --train_path ${data_path} \
-        --transform_path ${data_path} \
-        --report_path "report_logistic_regression_${cluster_type}.csv" \
-        $common_confs $spark_rapids_confs \
-        ${EXTRA_ARGS}
+    for num_classes in ${num_classes_list}; do
+
+        data_path=${gen_data_root}/classification/r${num_rows}_c${num_cols}_float32_ncls${num_classes}.parquet
```
eordentlich, line 369: Can you change random forest classifier to reuse the same num_class=2 dataset?

lijinf2:
build

python/run_benchmark.sh
lijinf2: Done. The num_classes=2 dataset is now reused.

eordentlich (approved):
Nice work!

PR #462, opened 2023-10-05
correct scaling for getting average norm 1

eordentlich:
scaling was fine for test purposes, but was wrong in the context of the preceding comment


eordentlich:
build

leewyang (approved):

PR #464, opened 2023-10-07
load data before comms and nccl abort to solve issues with empty partitions

eordentlich:
In further testing of handling of empty partitions added previously on workstation and dgx (local mode) found that even though we now raise an exception in this case, this can still result in hangs in python workers.  Sometimes the job hangs (dgx) and other times exits but the python worker isn't killed (docker on local workstation).    

Loading data before setting up comms (to prevent a worker with data from racing ahead and starting e.g an allreduce) and the nccl abort seems to clean things up in these cases.

I don't think this is the root cause of https://github.com/NVIDIA/spark-rapids-ml/issues/453 but should be addressed.

### DIFF
--- a/python/tests/test_nearest_neighbors.py
+++ b/python/tests/test_nearest_neighbors.py
@@ -326,7 +326,7 @@ def test_nearest_neighbors(
         random_state=0,
     )  # make_blobs creates a random dataset of isotropic gaussian blobs.
 
-    # set average norm to be 1 to allow comparisons with default error thresholds
+    # set average norm sq to be 1 to allow comparisons with default error thresholds
     # below
     root_ave_norm_sq = np.sqrt(np.average(np.linalg.norm(X, ord=2, axis=1) ** 2))
     X = X / root_ave_norm_sq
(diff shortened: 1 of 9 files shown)


eordentlich:
build

python/tests/test_nearest_neighbors.py
```diff
@@ -326,7 +326,7 @@ def test_nearest_neighbors(
         random_state=0,
     )  # make_blobs creates a random dataset of isotropic gaussian blobs.
 
-    # set average norm to be 1 to allow comparisons with default error thresholds
+    # set average norm sq to be 1 to allow comparisons with default error thresholds
```
eordentlich, line 329: This was still wrong so correcting in this PR.

leewyang (dismissed):

eordentlich:
Did some benchmarking with 2 gpus and the change (additional barrier) seems to add 1sec of latency to fit.

eordentlich:
build

eordentlich:
Moving data loading before nccl comms setup seems to also solve the issues without the extra barrier in previous version.   Tested with uvm on databricks and seems to be ok.
While testing had to upgrade spark-rapids jar to 23.08.2 to avoid databricks crashes.

leewyang (approved):

PR #466, opened 2023-10-10
changes to databricks init script for updated kernel drivers

eordentlich:
Databricks AWS recently updated gpu kernel driver versions.
Forward compatibility no longer needed and can cause problems in some runtimes.

Databricks Azure still has older drivers, so leaving needed lines in comments for notebook examples.


eordentlich:
NO CI needed.

leewyang (approved):
Could maybe add an argument/env-var with conditional statements, but probably not worth the effort, esp. if Azure Databricks updates eventually.

PR #475, opened 2023-10-20
improve docs by removing private methods, update logistic doc string …

eordentlich:
…and update package version

adding underscore prefix to some method names to avoid exposing in public api docs


eordentlich:
build

eordentlich:
build

leewyang (approved):

PR #478, opened 2023-10-23
adds a nightly update of draft release api docs

eordentlich:
for the WIP release

Also removing page source link from sphinx generated pages.   These were broken anyway due to jekyll underscore ignore, but don't think we really need them.


eordentlich:
build

leewyang (approved):

PR #494, opened 2023-10-31
[Doc] add performance tuning page

wbo4958:


docs/site/performance.md
```diff
@@ -0,0 +1,43 @@
...
+  --conf spark.task.resource.gpu.amount=0.08 \
+  ...
+```
+
+The above submit command specifies a request for 1 GPU and 12 CPUs per executor. So you can see,
+a total of 12 tasks per executor will be executed concurrently during the ETL phase.
+
+However, if you are using a spark-rapids-ml version earlier than 23.10.0 or a Spark
+standalone cluster version below 3.4.0, you still need to set
+`"spark.task.resource.gpu.amount"` equal to `"spark.executor.resource.gpu.amount"`. For example,
```
eordentlich, line 33: Why do they have to be equal?   We just need to make sure there is one task per gpu not necessarily one task per executor.

docs/site/performance.md
```diff
@@ -0,0 +1,43 @@
...
+spark-submit \
+  --master spark://<master-ip>:7077 \
+  --conf spark.executor.cores=12 \
+  --conf spark.task.cpus=1 \
+  --conf spark.executor.resource.gpu.amount=1 \
+  --conf spark.task.resource.gpu.amount=0.08 \
+  ...
+```
+
+The above submit command specifies a request for 1 GPU and 12 CPUs per executor. So you can see,
```
eordentlich, line 28: submit -> spark-submit

docs/site/performance.md
```diff
@@ -0,0 +1,43 @@
...
+a total of 12 tasks per executor will be executed concurrently during the ETL phase.
```
eordentlich, line 29: Maybe add that stage level scheduling is then used internal to the library to automatically carry out the ML training phases using the required 1 gpu per task.

docs/site/performance.md
wbo4958: Done

docs/site/performance.md
wbo4958: Yeah, You're right. New commit has fixed this part.

docs/site/performance.md
wbo4958: Done.

eordentlich (approved):
👍

eordentlich:
@wbo4958  Can this be merged to 23.10 also?

PR #497, opened 2023-11-01
Add test for stage-level scheduling

wbo4958:


wbo4958:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -696,60 +696,74 @@ def fitMultipleModels() -> List["_CumlModel"]:
...
+            # there will be only 1 task running at any time.
+            self.logger.info(
+                "Stage-level scheduling in spark-rapids-ml requires spark.executor.cores > 1 "
             )
-            return rdd
+            return True
 
-        if int(executor_gpu_amount) > 1:
-            # For spark.executor.resource.gpu.amount>1, we suppose user knows how to configure
+        if int(executor_gpus) > 1:
```
eordentlich, line 732: I don't think spark rapids supports executor_gpus > 1.   But shouldn't we allow this case for stage level scheduling without spark rapids for etl?

eordentlich (commented):
Looks good.   One comment/question.

python/src/spark_rapids_ml/core.py
```diff
@@ -696,60 +696,74 @@ def fitMultipleModels() -> List["_CumlModel"]:
         else:
             return super().fitMultiple(dataset, paramMaps)
 
-    def _try_stage_level_scheduling(self, rdd: RDD) -> RDD:
-        ss = _get_spark_session()
-        sc = ss.sparkContext
+    def _skip_stage_level_scheduling(self, spark_version: str, conf: SparkConf) -> bool:
```
leewyang, line 699: Any reason to ask the caller to provide the version and conf vs. just passing the sparkContext?  Seems weird/inefficient.  Saw that this is used _somewhat_ in tests, so maybe find a way to mock the version instead for testing?

python/src/spark_rapids_ml/core.py
wbo4958: For now, we can't do anything for this case because of the known spark bug.

python/src/spark_rapids_ml/core.py
wbo4958: yeah. It's for testing purposes.

eordentlich (approved):
👍

PR #498, opened 2023-11-01
output raw prediction column

eordentlich:
For logistic regression using api found by @lijinf2 https://docs.rapids.ai/api/cuml/stable/api/#cuml.LogisticRegression.decision_function which matches spark's actual raw prediction output.    probability and prediction columns derived using cupy from this column to avoid repeated inference computations.

For random forest, raw prediction col is set to a copy of probability col, for compatibility with Spark's BinaryClassificationEvaluator default columns.


eordentlich:
build

eordentlich:
build

leewyang (approved):

lijinf2 (approved):
LGTM

PR #499, opened 2023-11-02
initialize rmm with device for managed memory

eordentlich:
Fixes multi-gpu workers and executors when running cuda managed memory (uvm) and over subscribed gpu memory.


eordentlich:
build

leewyang (approved):

PR #501, opened 2023-11-02
Keep docstring consistent on supporting vector, array, and columnar input formats

lijinf2:
Removed featuresCols and inputCols in docstring as they have been included in featuresCol and inputCol.

### DIFF
--- a/python/src/spark_rapids_ml/clustering.py
+++ b/python/src/spark_rapids_ml/clustering.py
@@ -104,7 +104,7 @@ class _KMeansCumlParams(_CumlParams, _KMeansParams, HasFeaturesCols):
 
     def setFeaturesCol(self: P, value: Union[str, List[str]]) -> P:
         """
-        Sets the value of :py:attr:`featuresCol` or :py:attr:`featuresCols`. Used when input vectors are stored in a single column.
+        Sets the value of :py:attr:`featuresCol` or :py:attr:`featuresCols`.
         """
         if isinstance(value, str):
             self._set_params(featuresCol=value)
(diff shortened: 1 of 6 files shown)


lijinf2:
build

python/src/spark_rapids_ml/clustering.py
```diff
@@ -149,11 +149,10 @@ class KMeans(KMeansClass, _CumlEstimator, _KMeansCumlParams):
     tol: float (default = 1e-4)
         early stopping criterion if centers do not change much after an iteration.
 
-    featuresCol: str
-        the name of the column that contains input vectors. featuresCol should be set when input vectors are stored in a single column of a dataframe.
-
-    featuresCols: List[str]
```
eordentlich, line 155: I see a Param with this name and a mixin (HasFeaturesCols) as well as getters and setters.   But it is not included as a constructor arg.   So are we getting rid of featuresCols throughout?   Or should we add as a constructor param.  @wbo4958 ?

leewyang: FWIW, I think this was originally trying to mirror the `HasInputCol` and `HasInputCols` [params](https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/param/shared.html), but it is obviously redundant if `featuresCol` support both single str and list of str (which is a bit non-standard).  Anyhow, as long as we maintain compatibility w/ Pyspark code/usage, I think it's probably OK to remove.

wbo4958: Seems we're using featuresCols to store the feature names, seems we can't remove it.

wbo4958:
LGTM.

eordentlich (approved):
👍

PR #521, opened 2023-11-13
make tests compatible with spark 3.3

eordentlich:
and run fast tests nightly against 3.3

tested locally against both 3.3 and 3.4


eordentlich:
build

leewyang (approved):

PR #538, opened 2023-12-30
Support sparse vectors in LogisticRegression (Fit only)

lijinf2:


lijinf2:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -313,7 +405,7 @@ def _pyspark_class(self) -> Optional[ABCMeta]:
         )
 
 
-class _CumlCaller(_CumlParams, _CumlCommon):
+class _CumlCaller(HasEnableSparseDataOptim, _CumlParams, _CumlCommon):
```
eordentlich, line 408: This suggests we support sparse vectors in all our algos, while only logistic regression has support.   Seems it will break if sparse vectors are fed into other algos.

python/src/spark_rapids_ml/core.py
```diff
@@ -552,11 +663,15 @@ def _call_cuml_fit_func(
         array_order = self._fit_array_order()
 
         cuml_verbose = self.cuml_params.get("verbose", False)
+        use_sparse_array = (
+            alias.featureVectorType in dataset.schema.fieldNames()
```
eordentlich, line 667: Does this give a false positive if there is a column with the name alias.featureVectorType?   Maybe check specifically that the feature column is of struct type.

python/src/spark_rapids_ml/core.py
```diff
@@ -135,6 +153,80 @@
...
+            # Note: According to spark ML VectorUDT format,
+            # when type field is 1, the size field is also empty.
+            # we need to check the values field to get vector length.
+            vec_size = len(vec_values)
+            csr_indices = np.arange(vec_size, dtype=np.int32)
+            csr_values = vec_values
+
+        if n_features == 0:
+            n_features = vec_size
+        assert n_features == vec_size
```
eordentlich, line 216: Better to give more informative error message if this is not the case.  i.e all vectors must be of the same dimension.

python/src/spark_rapids_ml/core.py
```diff
@@ -135,6 +153,80 @@
...
+
+        csr_indices_list.append(csr_indices)
+        csr_indptr_list.append(csr_indptr_list[-1] + len(csr_indices))
```
eordentlich, line 219: Probably correct, but just checking it is expected that the length of this array is 1 more than the number of sparse rows.  ie. the last entry points to the start of the next non-existent row's index data.

python/src/spark_rapids_ml/params.py
```diff
@@ -39,6 +39,28 @@
...
+    This is a Params based class inherited from XGBOOST: https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/spark/params.py.
+    It holds the variable to store the boolean config of enabling sparse data optimization.
+    """
+
+    enable_sparse_data_optim = Param(
+        Params._dummy(),
+        "enable_sparse_data_optim",
+        "This stores the boolean config of enabling sparse data optimization, if enabled, "
+        "Spark rapids ml will construct a sparse matrix in the the format of csr_matrix, "
+        "then calls cuml with the sparse matrix. This config is disabled by default. If most of "
```
eordentlich, line 54: Technically it is enabled by default as the default None value bases sparse vs dense off first row.

python/src/spark_rapids_ml/utils.py
```diff
@@ -444,3 +447,23 @@ def translate_trees(sc: SparkContext, impurity: str, model: Dict[str, Any]):  #
...
+    except ImportError:
+        pass
+
+    try:
+        from pyspark.databricks.sql.functions import unwrap_udt as databricks_unwrap_udt
+
+        return databricks_unwrap_udt
+    except ImportError as exc:
+        raise RuntimeError(
+            "Cannot import pyspark `unwrap_udt` function. Please install pyspark>=3.4 "
```
eordentlich, line 467: So thinking about our spark 3.3 test.   Let's document that this is only supported in spark >= 3.4 and disable those tests.

python/src/spark_rapids_ml/classification.py
```diff
@@ -975,6 +989,7 @@ def _single_fit(init_parameters: Dict[str, Any]) -> Dict[str, Any]:
                     "n_cols": logistic_regression.n_cols,
                     "dtype": logistic_regression.dtype.name,
                     "num_iters": logistic_regression.solver_model.num_iters,
+                    "objective": logistic_regression.solver_model.objective,
```
eordentlich, line 992: Nice.   We can improve summary output with this.  I guess it is just final value not full history.

python/tests/test_logistic_regression.py
```diff
@@ -1257,3 +1276,258 @@ def test_functor(y: np.ndarray, err_msg: str) -> None:
...
+    fit_intercept: bool,
+    caplog: LogCaptureFixture,
+) -> None:
+    datatype = np.float32
+    tolerance = 0.001
+    reg_param = 1e-6
+
+    from scipy.sparse import csr_matrix
+    from sklearn.datasets import fetch_20newsgroups
+    from sklearn.feature_extraction.text import CountVectorizer
```
eordentlich, line 1426: Would be nicer to use pyspark versions of CountVectorizer and train_test_split to add another example of api compatibility.

lijinf2:
build

python/tests/test_logistic_regression.py
eordentlich: We should use this example to add sparse version of our example jupyter notebook: https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.12/notebooks/logistic-regression.ipynb

lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -931,8 +939,14 @@ def _logistic_regression_fit(
                 concated = pd.concat(X_list)
                 concated_y = pd.concat(y_list)
             else:
-                # features are either cp or np arrays here
-                concated = _concat_and_free(X_list, order=array_order)
+                if isinstance(X_list[0], scipy.sparse.csr_matrix):
+                    concated = scipy.sparse.vstack(X_list)
+                elif isinstance(X_list[0], cupyx.scipy.sparse.csr_matrix):
+                    concated = cupyx.scipy.sparse.vstack(X_list)
```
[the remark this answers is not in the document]
lijinf2: Done.

python/src/spark_rapids_ml/core.py
```diff
@@ -135,6 +153,80 @@
...
+    unwrap_udt = _get_unwrap_udt_fn()
+    features_unwrapped_vec_col = unwrap_udt(feature_col)
+
+    # After a `pyspark.ml.linalg.VectorUDT` type column being unwrapped, it becomes
+    # a pyspark struct type column, the struct fields are:
+    #  - `type`: byte
+    #  - `size`: int
+    #  - `indices`: array<int>
+    #  - `values`: array<double>
+    # For sparse vector, `type` field is 0, `size` field means vector length,
```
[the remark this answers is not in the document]
lijinf2: Revised.

python/src/spark_rapids_ml/core.py
```diff
@@ -379,17 +472,35 @@ def _pre_process_data(
...
+                        vector_to_array(col(input_col), vector_element_type).alias(alias.data)  # type: ignore
+                    )
+
                 if not self._float32_inputs:
                     feature_type = DoubleType
             else:
                 raise ValueError("Unsupported input type.")
 
-            dimension = len(dataset.first()[input_col])  # type: ignore
+            dimension = len(first_record[input_col])  # type: ignore
```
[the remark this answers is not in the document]
lijinf2: Yeah, see below copied from pyspark shell

>>> a = Vectors.sparse(10, [1., 2.], [1, 8])
>>> a
SparseVector(10, {1: 1.0, 2: 8.0})
>>> len(a)
10

python/src/spark_rapids_ml/core.py
lijinf2: Right. Moved the param to the algorithm class LogisticRegression.

python/src/spark_rapids_ml/core.py
lijinf2: It seems fine because core.py converts the feature column and label column into alias columns, through dataset = dataset.select(*select_cols). The select_cols all use alias names. 

Revised the code to also check the existence of alias.featureVectorSize and alias.featureVectorIndices, given the VectorUDT features column will be unwrapped into four alias columns (i.e. featureVectorType, featureVectorSize, featureVectorIndices, and data).

python/src/spark_rapids_ml/core.py
lijinf2: Added.

python/src/spark_rapids_ml/core.py
lijinf2: Added the checking of len(csr_indptr_list).

Did not add checking of last entry. It requires maintaining a counter to track the current number of non-zero values. Seems not checking is fine because the csr_matrix constructor used after will do some checking. Let me know if the checking is still favorable.

python/src/spark_rapids_ml/params.py
lijinf2: Yeah, revised the docstring to be more accurate.

python/src/spark_rapids_ml/utils.py
lijinf2: Yeah, added the checking.

python/src/spark_rapids_ml/classification.py
lijinf2: Right, cuml currently only returns the final value. Need to update C++ side if to match the full summary

python/tests/test_logistic_regression.py
lijinf2: Replaced the sklearn CountVectorizer with Spark CountVectorizer and RegexTokenizer. 

Thinking to add the example to notebook in the next PR. It seems the NLP20NEWS accuracy and training iterations are sensitive to the tokenizer and hyperparameters (i.e. RegParam and maxIter). May need to spend sometime to understand the behavior.

python/src/spark_rapids_ml/core.py
```diff
@@ -135,6 +153,84 @@
...
+
+
+# similar to the XGBOOST _read_csr_matrix_from_unwrapped_spark_vec in https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/spark/data.py
+def _read_csr_matrix_from_unwrapped_spark_vec(part: pd.DataFrame) -> csr_matrix:
+    # variables for constructing csr_matrix
+    csr_indices_list, csr_indptr_list, csr_values_list = [], [0], []
+
+    n_features = 0
+
+    for vec_type, vec_size_, vec_indices, vec_values in zip(
```
eordentlich, line 194: I wonder if there is a more efficient 'vectorized' approach to doing this.   Iterating in python can be slow.  Maybe add a TBD comment on this so we don't forget.

python/tests/test_logistic_regression.py
eordentlich: Sure.  Another PR for notebook is fine.

python/src/spark_rapids_ml/core.py
lijinf2: Yeah, added TBD comment.

eordentlich:
I almost forgot about it, but what about sparse transform?   This PR seems to handle fit only.

lijinf2:
build

lijinf2:
> I almost forgot about it, but what about sparse transform? This PR seems to handle fit only.

Good catch. the PR forgets about transform. It runs but seems to convert everything to dense arrays. Will update.

eordentlich (approved):
Transform support can be added in a separate PR.

PR #545, opened 2024-01-04
[LogisticRegression] Support sparse matrix in transform

lijinf2:


lijinf2:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -1436,7 +1482,9 @@ def predict_udf(iterator: Iterator[pd.DataFrame]) -> Iterator[pd.Series]:
                 cuml_objects[0] if isinstance(cuml_objects, list) else cuml_objects
             )
             for pdf in iterator:
-                if not input_is_multi_cols:
+                if use_sparse_array:
```
eordentlich, line 1485: Looks good.   I think you just need to make this change also where transform evaluate invokes cuml and should be good.  Here I think https://github.com/NVIDIA/spark-rapids-ml/blob/branch-23.12/python/src/spark_rapids_ml/core.py#L1298

eordentlich: And add some tests.

lijinf2:
build

python/src/spark_rapids_ml/core.py
lijinf2: Yeah, revised the code and tested through cross validator on sparse datasets.

eordentlich (approved):
👍

eordentlich:
Would be great to get an example notebook for the nlp example dataset from the test.

PR #548, opened 2024-01-06
skip testing sparse vector cross validator for spark < 3.4

lijinf2:
sparse vectors component uses a function that is only available in spark >=3.4, while nightly ci will test spark 3.3


lijinf2:
build

eordentlich (approved):
👍

PR #565, opened 2024-02-14
[LogisticRegression] Support standardization for dense vectors

lijinf2:
Sparse vectors will be densified to dense vectors if standardization is on. 

The PR removes uvm in the test of nlp20news dataset, because it is found the uvm leads to a cuda memset error in vars calculation.

### DIFF
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -660,7 +660,7 @@ class LogisticRegressionClass(_CumlClass):
             "fitIntercept": "fit_intercept",
             "threshold": None,
             "thresholds": None,
-            "standardization": "",  # Set to "" instead of None because cuml defaults to standardization = False
+            "standardization": "standardization",
             "weightCol": None,
             "aggregationDepth": None,
             "family": "",  # family can be 'auto', 'binomial' or 'multinomial', cuml automatically detects num_classes
@@ -680,6 +680,7 @@ class LogisticRegressionClass(_CumlClass):
     def _get_cuml_params_default(self) -> Dict[str, Any]:
         return {
             "fit_intercept": True,
+            "standardization": False,
             "verbose": False,
             "C": 1.0,
             "penalty": "l2",
@@ -848,6 +849,8 @@ class LogisticRegression(
         Note this is only supported in spark >= 3.4.
     fitIntercept:
         Whether to fit an intercept term.
+    standardization:
+        Whether to standardize the training data.
     num_workers:
         Number of cuML workers, where each cuML worker corresponds to one Spark task
         running on one GPU. If not set, spark-rapids-ml tries to infer the number of
@@ -909,6 +912,7 @@ class LogisticRegression(
         elasticNetParam: float = 0.0,
         tol: float = 1e-6,
         fitIntercept: bool = True,
+        standardization: bool = True,
         enable_sparse_data_optim: Optional[bool] = None,
         num_workers: Optional[int] = None,
         verbose: Union[int, bool] = False,
@@ -985,9 +989,18 @@ class LogisticRegression(
                     pdesc.rank,
                 )
 
+                intercept_array = logistic_regression.intercept_
+                # follow Spark to center the intercepts for multinomial classification
+                if (
+                    init_parameters["fit_intercept"] is True
+                    and len(intercept_array) > 1
+                ):
+                    intercept_mean = sum(intercept_array) / len(intercept_array)
+                    intercept_array -= intercept_mean
+
                 model = {
                     "coef_": logistic_regression.coef_.tolist(),
-                    "intercept_": logistic_regression.intercept_.tolist(),
+                    "intercept_": intercept_array.tolist(),
                     "classes_": logistic_regression.classes_.tolist(),
                     "n_cols": logistic_regression.n_cols,
                     "dtype": logistic_regression.dtype.name,
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -850,8 +850,8 @@ class LogisticRegression(
     fitIntercept:
         Whether to fit an intercept term.
     standardization:
-        Whether to standardize the training data. If true, spark rapids ml sets enable_sparse_data_optim=False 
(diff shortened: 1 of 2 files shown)


lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -952,8 +971,17 @@ def _logistic_regression_fit(
                 concated = _concat_and_free(X_list, order=array_order)
                 concated_y = _concat_and_free(y_list, order=array_order)
 
+            is_sparse = isinstance(concated, scipy.sparse.csr_matrix) or isinstance(
```
eordentlich, line 974: Any reason one would be sparse and the other dense?

python/src/spark_rapids_ml/classification.py
```diff
@@ -985,11 +1013,21 @@ def _single_fit(init_parameters: Dict[str, Any]) -> Dict[str, Any]:
                     pdesc.rank,
                 )
 
+                intercept_array = logistic_regression.intercept_
+                # follow Spark to center the intercepts for multinomial classification
+                if (
+                    init_parameters["fit_intercept"] is True
+                    and len(intercept_array) > 1
+                ):
+                    intercept_mean = sum(intercept_array) / len(intercept_array)
```
eordentlich, line 1022: Is there a 'mean' method that can be called?   Also, how does this not change the model output?

python/tests/test_logistic_regression.py
```diff
@@ -291,8 +304,13 @@ def test_classifier(
     reg_param: float = 0.0,
     elasticNet_param: float = 0.0,
     tolerance: float = 0.001,
+    standardization: bool = False,
     convert_to_sparse: bool = False,
 ) -> LogisticRegression:
+    assert (
+        standardization is False
+    ), "standardization=True is not supported due to testing with single-GPU LogisticRegression"
```
eordentlich, line 312: Can we use cuml standard scalar preprocessr?

python/tests/test_logistic_regression.py
```diff
@@ -463,13 +482,9 @@ def test_compat(
...
 
         assert _LogisticRegression().getRegParam() == 0.0
-        if lr_types[0] is SparkLogisticRegression:
-            blor = _LogisticRegression(
-                regParam=0.1, fitIntercept=fit_intercept, standardization=False
-            )
-        else:
-            warnings.warn("spark rapids ml does not accept standardization")
-            blor = _LogisticRegression(regParam=0.1, fitIntercept=fit_intercept)
+        blor = _LogisticRegression(
```
eordentlich, line 485: Should we now have a version of this test with standardization on?

python/tests/test_logistic_regression.py
```diff
@@ -109,7 +112,9 @@ def test_toy_example(gpu_number: int) -> None:
         schema = features_col + " array<float>, " + label_col + " float"
         df = spark.createDataFrame(data, schema=schema)
 
-        lr_estimator = LogisticRegression(regParam=1.0, num_workers=gpu_number)
+        lr_estimator = LogisticRegression(
+            standardization=False, regParam=1.0, num_workers=gpu_number
```
eordentlich, line 116: Do we have a test comparing objective or coefficients to cpu spark that would have failed before given that spark has standardization on by default?

python/src/spark_rapids_ml/classification.py
lijinf2: check if concated is sparse, if so and if standardization is on, it will be densified in order to use standardization.

python/src/spark_rapids_ml/classification.py
lijinf2: Yeah, revised the code to use np.mean and cp.mean

python/tests/test_logistic_regression.py
lijinf2: revised the code to remove the standardization option from argument list. 
this test was for standardization=False only.

python/tests/test_logistic_regression.py
lijinf2: Yeah, revised to add standardization=True

python/tests/test_logistic_regression.py
lijinf2: Yes, the PR compares coefficients. There is a compare_model to assert CPU and GPU having same coefficients. The compare_model is used everywhere in the test cases of this PR when testing standardization=True.

The test_compat function that was just revised shows the coefficients are different when standardization is set to True and False. GPU and CPU match in both cases.

Regarding objective, the test_sparse_nlp20news compares that in both cases when standardization = True and False.

lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -985,11 +1013,27 @@ def _single_fit(init_parameters: Dict[str, Any]) -> Dict[str, Any]:
...
+                    init_parameters["fit_intercept"] is True
+                    and len(intercept_array) > 1
+                ):
+                    import cupy as cp
+
+                    intercept_mean = (
+                        np.mean(intercept_array)
+                        if isinstance(intercept_array, np.ndarray)
+                        else cp.mean(intercept_array)
+                    )
```
eordentlich, line 1028: Doesn't this impact the model prediction ?

eordentlich (commented):
Nice.   Just one more question.

python/src/spark_rapids_ml/classification.py
lijinf2: Good question. It seems not affect the predicted class. Logistic is monotonically increasing and softmax predicts to the class that has the largest logistic value.

lijinf2: It seems affect the rawPredict value but not max(rawPredict values)

lijinf2: Reference in Spark code: https://github.com/apache/spark/blob/2a7e8900ab5f3845c3ebe6f43ce9b5ccd95cec99/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala#L682

eordentlich: I see.  It is only for multi-class and the soft-max is unaffected since each exponential is scaled by the same factor.

eordentlich (approved):
👍

PR #569, opened 2024-02-20
[LogisticRegression] Padding zeros to bypass the out-of-bound memory access of the raft stats (sum, mean, stddev) kernels

lijinf2:
This PR is a workaround to bypass gpu memory out-of-bound access of raft 24.02.

### DIFF
--- a/python/src/spark_rapids_ml/classification.py
+++ b/python/src/spark_rapids_ml/classification.py
@@ -979,6 +979,9 @@ class LogisticRegression(
             if self.getStandardization() is True and is_sparse is True:
                 concated = concated.toarray()
 
+            # Padding zero columns so that the total number of columns is a multiple of 32.
+            # The purpose is to avoid out-of-bound memory access of some RAFT 24.02 cuda kernels (e.g. sum).
+            # This is a temporary workaround and is expected to be removed in 24.04
             num_pad_zero_cols = 0
             if self.getStandardization() is True and concated.shape[1] % 32 != 0:
                 num_pad_zero_cols = 32 - concated.shape[1] % 32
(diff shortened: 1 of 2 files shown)


lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -979,9 +979,33 @@ def _logistic_regression_fit(
             if self.getStandardization() is True and is_sparse is True:
                 concated = concated.toarray()
 
+            num_pad_zero_cols = 0
+            if self.getStandardization() is True and concated.shape[1] % 32 != 0:
+                num_pad_zero_cols = concated.shape[1] % 32
```
eordentlich, line 984: Shouldn't the number of padding columns be 32 - concated.shape[1] % 32 ?

lijinf2:
build

python/src/spark_rapids_ml/classification.py
lijinf2: Good catch!

python/src/spark_rapids_ml/classification.py
```diff
@@ -979,9 +979,35 @@ def _logistic_regression_fit(
             if self.getStandardization() is True and is_sparse is True:
                 concated = concated.toarray()
 
+            num_pad_zero_cols = 0
```
eordentlich, line 982: Maybe add a comment about why this is done and that it is temporary work around.

eordentlich (commented):
Looks good.  One more minor comment.

lijinf2:
build

python/src/spark_rapids_ml/classification.py
lijinf2: Sounds good. Added a comment.

lijinf2:
build

eordentlich (approved):
👍

PR #574, opened 2024-02-23
Multinomial Logistic Regression datagen and Column-specific density support for sparse regression datagen

Er1cCheng:


Er1cCheng:
build

python/benchmark/test_gen_data.py
```diff
@@ -159,16 +163,43 @@ def test_make_regression(
...
+            assert c.shape == (col_num,), "coef shape mismatch"
+            assert np.count_nonzero(c) == 3, "Unexpected number of informative features"
+        else:
+            print(
+                c.shape,
+                (
+                    n_classes,
+                    col_num,
+                ),
+            )
```
lijinf2, line 184: seems we can remove print statement. machine will not check it.

python/benchmark/test_gen_data.py
```diff
@@ -159,16 +163,43 @@ def test_make_regression(
...
+            # Test that y consists of only discrete label
+            possible_labels = range(n_classes_num)
             for n in y:
-                assert n == 0 or n == 1
+                found = False
+                for l in possible_labels:
+                    if n == l:
+                        found = True
+                        break
+                assert found, "Invalid label"
```
lijinf2, line 202: Have we already checked unique(y) == n_classes_num?

python/benchmark/test_gen_data.py
```diff
@@ -177,17 +208,29 @@ def test_make_regression(
 @pytest.mark.parametrize("dtype", ["float64"])
 @pytest.mark.parametrize("use_gpu", ["True", "False"])
 @pytest.mark.parametrize("redundant_cols", ["0", "2"])
```
lijinf2, line 210: thinking to mark some parameter values slow.

python/benchmark/test_gen_data.py
```diff
@@ -258,7 +329,7 @@ def test_make_sparse_regression(
 
         # If there is no random shuffled redundant cols, we can check the total density
         density_num = float(density)
-        if redundant_cols == "0":
+        if redundant_cols == "0" and density_curve == "None":
```
lijinf2, line 332: Seems only density_curve="None" is tested. How about density_curve="Linear" and "Exponential"?

python/benchmark/test_gen_data.py
```diff
@@ -238,15 +286,38 @@ def test_make_sparse_regression(
...
+            assert c.shape == (col_num,), "coef shape mismatch"
+            assert np.count_nonzero(c) == 3, "Unexpected number of informative features"
+        else:
+            print(
+                c.shape,
+                (
+                    n_classes,
+                    col_num,
+                ),
+            )
```
lijinf2, line 300: delete print statement

python/benchmark/test_gen_data.py
```diff
@@ -238,15 +286,38 @@ def test_make_sparse_regression(
...
+            # Test that X consists of only discrete label
+            possible_labels = range(n_classes_num)
             for n in y:
-                assert n == 0 or n == 1
+                found = False
+                for l in possible_labels:
+                    if n == l:
+                        found = True
+                        break
+                assert found, "Invalid label"
```
lijinf2, line 320: Can we check unique(y) == n_classes

python/benchmark/gen_data.py
```diff
@@ -158,6 +158,8 @@ def _add_extra_parameters(self) -> None:
                 help_msg = "boolean for whether the shuffle the rows and cols of the feature matrix"
             elif name == "tail_strength":
                 help_msg = "tail strength for random low rank feature matrix generation, refer to sklearn.datasets.make_low_rank_matrix()"
+            elif name == "density_curve":
+                help_msg = "Specify columns wise density curve, support Linear or Exponential. Argument density would not be used to represent the max density in the curve"
```
lijinf2, line 162: What does Linear and Exponential mean?

lijinf2: what will be the output if density_curve is set to Linear or Exponential?

python/benchmark/gen_data_distributed.py
```diff
@@ -695,26 +822,57 @@ def make_sparse_regression_udf(
...
-                    y = np.random.binomial(1, prob)
+                    if use_cupy:
+                        prob = 1 - 1 / (1 + cp.exp(-y_p))
+                        del y_p
+                        y = cp.random.binomial(1, prob)
+                    else:
+                        prob = 1 - 1 / (1 + np.exp(-y_p))
+                        del y_p
+                        y = np.random.binomial(1, prob)
+            else:
```
lijinf2, line 874: large chunk of code looks identical in dense class and sparse class. Can we merge identical code (e.g. create a python function)?

python/benchmark/test_gen_data.py
Er1cCheng: For linear and exponential, the thing is that after shuffling the matrix seems just purely random. Is there a good way to test it?

python/benchmark/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+    parser.addoption(
+        "--runslow", action="store_true", default=False, help="run slow tests"
+    )
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    config.addinivalue_line("markers", "slow: mark as slow to run")
+
+
+def pytest_collection_modifyitems(
```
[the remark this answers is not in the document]
Er1cCheng: The conftest file has to exist in the directory you want to control. This is a shrunked file compared to tests/conftest.py since we only need to support the slow option.

python/benchmark/test_gen_data.py
Er1cCheng: There is the possibility that during randomness we lose completely a class even if less likely given sufficient data samples.

python/benchmark/gen_data_distributed.py
Er1cCheng: Seems like an extra user defined function inside the UDF would break the pyspark dependency system. I get the same "Unable to find module gen_data_distribtued" error as last time.

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -299,6 +299,41 @@ def make_matrix_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+#         if use_cupy:
+#             prob = 1 - 1 / (1 + cp.exp(-y_p))
+#             del y_p
+#             y = cp.random.binomial(1, prob)
+#         else:
+#             prob = 1 - 1 / (1 + np.exp(-y_p))
+#             del y_p
+#             y = np.random.binomial(1, prob)
+
+#     return y
```
lijinf2, line 334: The comments should be deleted if not useful

Er1cCheng:
build

Er1cCheng:
build

python/benchmark/conftest.py
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2022, NVIDIA CORPORATION.
```
wbo4958, line 2: 2022 -> 2024

python/run_benchmark.sh
```diff
@@ -441,39 +441,42 @@ if [[ "${MODE}" =~ "logistic_regression" ]] || [[ "${MODE}" == "all" ]]; then
...
-            --num_cpus $num_cpus \
-            --num_runs $num_runs \
-            --train_path ${data_path} \
-            --transform_path ${data_path} \
-            --report_path "report_logistic_regression_${cluster_type}.csv" \
-            $common_confs $spark_rapids_confs \
-            ${EXTRA_ARGS}
+            if [[ ! -d ${data_path} ]]; then
+                python $gen_data_script sparse_regression \
+                --n_informative $( expr $num_cols / 3 )  \
```
eordentlich, line 449: should we have a new parameter num_sparse_cols instead of num_cols here?   We might want to make this much larger than the other dense cases.

python/benchmark/gen_data_distributed.py
```diff
@@ -455,21 +487,63 @@ def make_regression_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+                    if multinomial_log:
+                        if use_cupy:
+                            y = y.get()
+
+                        probs = [
+                            sp.special.softmax(target_weight) for target_weight in y
+                        ]
+
+                        multi_labels = [
+                            random.choices(range(n_classes), weights=p)[0]
```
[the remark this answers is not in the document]
Er1cCheng: Can't really find a way to achieve the matrix version, the softmax one does have a axis option

Er1cCheng:
build

python/benchmark/gen_data.py
```diff
@@ -164,6 +164,13 @@ def _add_extra_parameters(self) -> None:
             else:
                 help_msg = ""
 
+            # Support multiple biases
+            if name == "bias":
```
eordentlich, line 168: Why not generate bias randomly like the weights?

Er1cCheng: This is code inherited from before. Do we want random bias then?

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

python/benchmark/test_gen_data.py
```diff
@@ -177,19 +219,49 @@ def test_make_regression(
...
+    [
+        ("None", "True"),
+        ("Linear", "True"),
+        ("Linear", "False"),
+        ("Exponential", "False"),
+        pytest.param("Exponential", "True", marks=pytest.mark.slow),
+    ],
+)
+@pytest.mark.parametrize(
+    "n_chunks", ["10", pytest.param("100", marks=pytest.mark.slow)]
```
lijinf2, line 249: what is the relation of density, density_curve, and chunk?

Er1cCheng:
build

python/benchmark/test_gen_data.py
```diff
@@ -257,13 +365,70 @@ def test_make_sparse_regression(
...
+                )
+                density_values *= n_chunks_num * density_num / sum(density_values)
+
+            col_per_chunk = np.full(n_chunks_num, orig_cols // n_chunks_num)
+            col_per_chunk[: (orig_cols % n_chunks_num)] += 1
+            col_per_chunk = np.cumsum(col_per_chunk)
+
+            density_idx = 0
+            dense_count = 0
+            for idx, col in enumerate(X_np):
```
lijinf2, line 409: what is the value of density_idx when the for loop is finished?

python/benchmark/test_gen_data.py
```diff
@@ -257,13 +365,70 @@ def test_make_sparse_regression(
...
+                # Ignore the redundant columns
+                if idx >= orig_cols:
+                    break
+
+                if density_values[density_idx] == idx:
```
lijinf2, line 414: not sure but it seems the if statement is always False? density_values[density_idx] is a float type, while idx is an int type?

Er1cCheng:
build

python/benchmark/test_gen_data.py
```diff
@@ -257,13 +365,70 @@ def test_make_sparse_regression(
...
+            num_partitions = 3
+
+            if isinstance(density, List):
+                density_num = float(density[0])
+            else:
+                density_num = float(density)
+
+            if density_curve == "Linear":
+                density_values = np.linspace(
+                    num_partitions / row_num, density_num, n_chunks_num
```
lijinf2, line 392: why density_values start from "num_partitions / row_num" not 0.? Can you give some explanation?

python/benchmark/test_gen_data.py
lijinf2: please add the checking of the value after for loop is finished.

python/benchmark/test_gen_data.py
Er1cCheng: The starting point for this is that I want to gaurantee the smallest density for each partition is big enough to generate at least one dense entry for the column. Thus, the density need to be at least 1/(rows/num_par) = num_par/rows. One potential issue is that num_partition is larger than row_num but that would defeat the whole purpose of this distributed datagen, so it should not happen at all. I wonder what's the "not 0." issue?

Er1cCheng:
build

lijinf2 (approved):
Looks nice!

PR #585, opened 2024-03-07
Support return VectorUDT in PCA transform

lijinf2:


lijinf2:
build

lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -1505,6 +1507,13 @@ def predict_udf(iterator: Iterator[pd.DataFrame]) -> Iterator[pd.Series]:
         pred_col = predict_udf(struct(*select_cols))
 
         if self._is_single_pred(dataset.schema):
+            if "array" in output_schema:
```
eordentlich, line 1510: Seems better to check the type of pred_col specifically since that is what will be transformed potentially.

python/tests/test_pca.py
```diff
@@ -379,7 +380,24 @@ def test_pca_spark_compat(
...
-        output = model.transform(df).collect()[0].output
+        k = model.getK()
+        output_df = model.transform(df)
+        schema_fields = [
+            StructField("features", VectorUDT(), True),
+        ]
+        if _PCA is SparkPCA:
+            schema_fields.append(
+                StructField(
+                    "output", VectorUDT(), True, metadata={"ml_attr": {"num_attrs": k}}
```
eordentlich, line 391: Is this new?   Should we match this?

python/src/spark_rapids_ml/core.py
lijinf2: Yeah, the type of pred_col is obtained through output_schema. Is there a way to get type from the pred_col Column variable?
 
Revised the code to handle the case when output_schema is StructType.

python/tests/test_pca.py
lijinf2: No idea whether the metadata is now and how it is used. 

Seems not necessary to match it for now.

lijinf2:
build

python/src/spark_rapids_ml/core.py
eordentlich: Can you look up the type of pred_col specifically in the output_schema vs just checking if there is an array type in output_schema?

lijinf2: The output_schema of PCA is "array<float>" without the column name. Seems there is no way to lookup through pred column name. Is there any other way?

lijinf2: Note `self._is_single_pred(dataset.schema)` is True so there is only one column in the output_schema. Seems ok to check the "array<float>" and "array<double>" keyword.

eordentlich (approved):
👍

PR #587, opened 2024-03-08
keep eval related computations and data on gpu

eordentlich:


wbo4958:
Overall, the idea looks good to me. However, seems we can do further enhancement if we ensure the predicted/transformed data is on the GPU before evaluating. like [here](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-24.02/python/src/spark_rapids_ml/regression.py#L746) and [here](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-24.02/python/src/spark_rapids_ml/classification.py#L1474-L1500)

eordentlich:
That is exactly what is being done in the PR here:  https://github.com/NVIDIA/spark-rapids-ml/pull/587/files#diff-5bffca677cda02544897eeac447779db91115ec59138bc10c6115e56ed123692R1476-R1482
But only for logistic regression so far.  Needs to be extended to the other crossvalidator supported models like random forest and linear regression.

eordentlich:
build

eordentlich:
build

python/src/spark_rapids_ml/regression.py
```diff
@@ -741,9 +746,16 @@ def _construct_lr() -> CumlT:
 
             return lrs
 
-        def _predict(lr: CumlT, pdf: TransformInputType) -> pd.Series:
-            ret = lr.predict(pdf)
-            return pd.Series(ret)
+        if eval_metric_info:
+
+            def _predict(lr: CumlT, pdf: TransformInputType) -> pd.Series:
```
lijinf2, line 751: nit: Is the return type CPU pd.Series or a GPU type (e.g. cp.ndarray)?

eordentlich: good catch, should be cp.ndarray

eordentlich:
build

lijinf2 (approved):

eordentlich:
build

PR #598, opened 2024-03-26
DBSCAN Algorithm for Spark RAPIDS ML

Er1cCheng:
Adding in the support and tests for DBSCAN clustering algorithm to Spark Rapids ML


python/tests/test_dbscan.py
```diff
@@ -0,0 +1,256 @@
...
+                data = tuple(row[0].toArray())
+            elif isinstance(row[0], np.float32) or isinstance(row[0], np.float64):
+                data = tuple(row)
+            else:
+                data = tuple(row[0])
+
+            label_rapids = label_arr[rid]
+            label_cuml = sample_to_cluster[data]
+
+            if label_rapids in cluster_dict:
```
lijinf2, line 253: Can we assert  "label_rapids == label_cuml" here?

lijinf2: Another way is to adopt cuml tests that assert the labels of core points are the same, like here: https://github.com/rapidsai/cuml/blob/2c876b50be7422857806baedabbf66b498b56e21/python/cuml/tests/dask/test_dask_dbscan.py#L233

We may add a unique id column to spark dataframe to leverage cuml_dbscan.core_sample_indices_.

python/tests/test_dbscan.py
```diff
@@ -0,0 +1,256 @@
...
+        [5, 2, 1, 3, 4],
+    ]
+
+    with CleanSparkSession() as spark:
+        feature_cols = ["c1", "c2", "c3", "c4", "c5"]
+        schema = ", ".join([f"{c} {data_type}" for c in feature_cols])
+        df = spark.createDataFrame(data, schema=schema)
+        dbscan = DBSCAN(num_workers=gpu_number, featuresCols=feature_cols)
+        dbscan_model = dbscan.fit(df)
+        label_df = dbscan_model.transform(df)
```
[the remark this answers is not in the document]
Er1cCheng: This test (referred from KMeans) meant to test that the algorithm works fine with different datatypes. There is no output checking the KMeans originally.

python/tests/test_dbscan.py
Er1cCheng: Assert "label_rapids == label_cuml" does work, but I thought it's safer to do this cluster mapping in case there is discrepency between the exact cluster label even if the algorithm is working correctly. Should we swap to the simple assert equal?

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+        self._set_params(idCol=value)
+        return self
+
+    def getIdCol(self) -> str:
+        """
+        Gets the value of `idCol`.
+        """
+        return self.getOrDefault("idCol")
+
+    def _ensureIdCol(self, df: DataFrame) -> DataFrame:
```
lijinf2, line 240: Can we move idCol, setter, getter, and _ensureIdCol to a new class and reuse the class in both NearestNeighbors and DBSCAN?

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+            * ``4 or False`` - Enables all messages up to and including information messages.
+            * ``5 or True`` - Enables all messages up to and including debug messages.
+            * ``6`` - Enables all messages up to and including trace messages.
+
+    max_mbytes_per_batch(optional): int
+        Calculate batch size using no more than this number of megabytes for the pairwise distance computation.
+        This enables the trade-off between runtime and memory usage for making the N^2 pairwise distance computations more tractable for large numbers of samples.
+        If you are experiencing out of memory errors when running DBSCAN, you can set this value based on the memory size of your device.
+
+    calc_core_sample_indices(optional): boolean (default = True)
```
lijinf2, line 307: If true, do we ever use the computed core indices? Seems computed only but never used.

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+    |[0.0, 0.0]|         0|
+    |[1.0, 1.0]|         0|
+    |[9.0, 8.0]|         1|
+    |[8.0, 9.0]|         1|
+    +----------+----------+
+    >>> gpu_dbscan.save("/tmp/dbscan")
+    >>> gpu_model.save("/tmp/dbscan_model")
+
+    >>> # vector column input
+    >>> from spark_rapids_ml.clustering import KMeans
```
lijinf2, line 349: import dbscan?

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+
+        self.verbose = verbose
+
+    def setEps(self: P, value: float) -> P:
+        return self._set_params(eps=value)
+
+    def getEps(self) -> float:
+        return self.getOrDefault("eps")
+
+    def setMin_samples(self: P, value: int) -> P:
```
lijinf2, line 407: Does spark have the underscore naming convention? or it expects setMinSamples

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,629 @@
...
+        return {
+            "eps": 0.5,
+            "min_samples": 5,
+            "metric": "euclidean",
+            "verbose": False,
+            "max_mbytes_per_batch": None,
+            "calc_core_sample_indices": True,
+        }
+
+    def _pyspark_class(self) -> Optional[ABCMeta]:
```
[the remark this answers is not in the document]
Er1cCheng: For returning the corresponding built-in spark ML class

python/src/spark_rapids_ml/dbscan.py
Er1cCheng: Yeah, but the default from cuML is True so I think this should be a match

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+            # experiments indicate it is faster to convert to numpy array and then to cupy array than directly
+            # invoking cupy array on the list
+            if cuda_managed_mem_enabled:
+                features = (
+                    cp.array(features)
+                    if self.use_sparse_array is False
+                    else cupyx.scipy.sparse.csr_matrix(features)
+                )
+
+            inputs.append(features)
```
lijinf2, line 609: code snippet for preparing inputs variable seemed very similar to that of core.py. Is it possible to reuse the code snippet of core.py? 

Seems you only need the concated variable in _cuml_fit, while you should be able to use dfs variable to construct concated.

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,747 @@
...
+                min_samples=self.getOrDefault("min_samples"),
+                metric=self.getOrDefault("metric"),
+                max_mbytes_per_batch=self.getOrDefault("max_mbytes_per_batch"),
+                calc_core_sample_indices=self.getOrDefault("calc_core_sample_indices"),
+                verbose=self.verbose,
+            )
+            dbscan.n_cols = params[param_alias.num_cols]
+            dbscan.dtype = np.dtype(dtype)
+
+            res = list(dbscan.fit_predict(concated).to_numpy())
```
lijinf2, line 642: Are we able to construct concated from dfs? If so, we can avoid the duplicating code snippet that construct concated from the dataframe.

python/tests/test_dbscan.py
```diff
@@ -0,0 +1,360 @@
...
+    for rid, row in enumerate(np_df):
+        label = cuml_transformed[rid]
+
+        sample_to_cluster[tuple(row)] = label
+
+    conf = {"spark.sql.execution.arrow.maxRecordsPerBatch": str(max_record_batch)}
+    with CleanSparkSession(conf) as spark:
+        df, features_col, _ = create_pyspark_dataframe(
+            spark, feature_type, data_type, X_dist, None
+        )
```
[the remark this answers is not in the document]
lijinf2: It seems not straightforward to keep the order of rows consistent in distance matrix and spark dataframe. Let's move the precomputed mode to the next PR.

python/src/spark_rapids_ml/dbscan.py
Er1cCheng: The issue here is that we need to work on the original full dataset instead of the dfs passed into _cuml_fit which is the partitioned dataframe for mapInPandas. There is no way to reconstruct from dfs the full dataset. The code is iterating on the broadcast varibales. The code snippet is partially duplicated due to it's a subportion of code from core.py -- _call_cuml_fit_func, while we have to extract this portion of code to work on a different set of broadcast variables (not the passed data to core.py).

python/tests/test_dbscan.py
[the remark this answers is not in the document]
Er1cCheng: Not doing recomputed anymore

Er1cCheng:
build

Er1cCheng:
build

python/tests/test_dbscan.py
eordentlich: Nice way to address possible permutation of labels.

python/tests/test_dbscan.py
```diff
@@ -0,0 +1,253 @@
...
+    with CleanSparkSession() as spark:
+        feature_cols = ["c1", "c2", "c3", "c4", "c5"]
+        schema = ", ".join([f"{c} {data_type}" for c in feature_cols])
+        df = spark.createDataFrame(data, schema=schema)
+        dbscan = DBSCAN(num_workers=gpu_number, featuresCols=feature_cols)
+        dbscan_model = dbscan.fit(df)
+        label_df = dbscan_model.transform(df)
+
+
+@pytest.mark.parametrize("feature_type", pyspark_supported_feature_types)
```
eordentlich, line 159: If sparse vectors are supported, we should have a test with that.

python/src/spark_rapids_ml/dbscan.py
eordentlich: Ok to diverge from cuml if no advantage to expose (say it is used internally to cuml testint).

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,655 @@
...
+
+        model._num_workers = self.num_workers
+        self._copyValues(model)
+
+        return model
+
+    def _create_pyspark_model(self, result: Row) -> _CumlModel:
+        raise NotImplementedError("DBSCAN does not support model creation from Row")
+
+    def _get_cuml_fit_func(
```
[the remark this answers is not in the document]
Er1cCheng: These abstract methods has to be implemented to get the class to work. The base is considered abstract.

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,655 @@
...
+        Optional[List[str]],
+        int,
+        Union[Type[FloatType], Type[DoubleType]],
+    ]:
+        (
+            select_cols,
+            multi_col_names,
+            dimension,
+            feature_type,
+        ) = _CumlCaller._pre_process_data(self, dataset)
```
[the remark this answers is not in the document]
Er1cCheng: This needs to specify the preprocess for Caller since the DBSCANModel is both a caller and a model, both of which contains a pre_process_data method (different ones)

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,655 @@
...
+        return _cuml_fit
+
+    def _get_cuml_transform_func(
+        self, dataset: DataFrame, eval_metric_info: Optional[EvalMetricInfo] = None
+    ) -> Tuple[
+        _ConstructFunc,
+        _TransformFunc,
+        Optional[_EvaluateFunc],
+    ]:
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
Er1cCheng: Same as above, need to implement in this class

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,655 @@
...
+        idCol_name = self.getIdCol()
+
+        default_num_partitions = dataset.rdd.getNumPartitions()
+
+        rdd = self._call_cuml_fit_func(
+            dataset=dataset,
+            partially_collect=False,
+            paramMaps=None,
+        )
+        rdd = rdd.repartition(default_num_partitions)
```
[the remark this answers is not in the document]
Er1cCheng: Yes, the rdd somehow has to have the same number of partition before and after the operations.

python/src/spark_rapids_ml/params.py
```diff
@@ -85,6 +87,46 @@ def getFeaturesCols(self) -> List[str]:
...
+        """
+        return self.getOrDefault("idCol")
+
+    def _ensureIdCol(self, df: DataFrame) -> DataFrame:
+        """
+        Ensure an id column exists in the input dataframe. Add the column if not exists.
+        """
+        if not self.isSet("idCol") and self.getIdCol() in df.columns:
+            raise ValueError(
+                f"Cannot create a default id column since a column with the default name '{self.getIdCol()}' already exists."
```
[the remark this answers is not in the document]
Er1cCheng: Fixed, the setter is still exposed to the user since if they have a dataframe with an id column they can just use it.

python/src/spark_rapids_ml/dbscan.py
```diff
@@ -0,0 +1,655 @@
...
+        inputs = []  # type: ignore
+
+        idCol = list(
+            self.idCols_[0].value
+            if len(self.idCols_) == 1
+            else np.concatenate([chunk.value for chunk in self.idCols_])
+        )
+
+        for pdf_bc in self.raw_data_:
+            pdf = pd.DataFrame(data=pdf_bc.value, columns=self.processed_input_cols)
```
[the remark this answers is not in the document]
Er1cCheng: This is to match and reutilize the data processing code after this line refered from core.py

Er1cCheng:
build

Er1cCheng:
build

eordentlich (approved):
LGTM.   Nice job 👍 

In your next PR, add DBSCAN to the supported algos in the top level README.

PR #607, opened 2024-04-08
Sparse Datagen Fix

Er1cCheng:
Fix sparse datagen bugs:

- Density curve values exceeding 1

- Tests not passing in n_chunk parameter

### DIFF
--- a/python/benchmark/test_gen_data.py
+++ b/python/benchmark/test_gen_data.py
@@ -297,6 +297,7 @@ def test_make_sparse_regression(
         n_chunks,
     ]
 
+    # Add parameters with multiple values
     input_args.append("--bias")
     if isinstance(bias, List):
         input_args.extend(bias)
--- a/python/benchmark/test_gen_data.py
+++ b/python/benchmark/test_gen_data.py
@@ -234,7 +234,7 @@ def test_make_regression(
 )
 @pytest.mark.parametrize(
     "density",
-    ["0.25", ["0.05", "0.1", "0.2"], pytest.param("0.2", marks=pytest.mark.slow)],
+    ["0.25", ["0.05", "0.1", "0.2"]],
 )
 @pytest.mark.parametrize(
     "rows, cols",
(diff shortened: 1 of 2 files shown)


Er1cCheng:
build

Er1cCheng:
build

Er1cCheng:
build

python/benchmark/test_gen_data.py
```diff
@@ -244,15 +244,12 @@ def test_make_regression(
...
-        ("Linear", "True"),
         ("Linear", "False"),
         ("Exponential", "False"),
         pytest.param("Exponential", "True", marks=pytest.mark.slow),
     ],
 )
-@pytest.mark.parametrize(
-    "n_chunks", ["10", pytest.param("100", marks=pytest.mark.slow)]
-)
+@pytest.mark.parametrize("n_chunks", ["100"])
```
eordentlich, line 252: Is this case not actually slow?   Same question for other change above.

python/benchmark/test_gen_data.py
```diff
@@ -416,9 +419,10 @@ def test_make_sparse_regression(
                 assert dense_count >= chunk_size * num_partitions * int(
                     (row_num // num_partitions) * col_density - 1
                 ) and dense_count <= chunk_size * num_partitions * int(
-                    (row_num // num_partitions) * col_density + 1
+                    (row_num // num_partitions + 1) * col_density + 1
```
eordentlich, line 422: Is this the actual guarantee on dense_count you are aiming for?

Er1cCheng: Yes, so for the number of rows in each partition it's either (row_num // numpartitions) or (row_num // numpartitions + 1). Then the upper bound should have a plus for corner case (Say that row_num mod num_partitions = num_partition - 1).

python/benchmark/test_gen_data.py
Er1cCheng: Was trying to reduce the number of tests to avoid too large testing runtime. A number greater than 10 would good since 10 is the default and we want to test for something more rigorous. Do you have any good number in mind (like 20 maybe?)

lijinf2 (approved):

PR #622, opened 2024-04-16
Algorithm parameter support for DBSCAN

Er1cCheng:
Add in algorithm parameter support for DBSCAN, could choose between "brute" and "rbc" algorithm

### DIFF
--- a/python/tests/test_dbscan.py
+++ b/python/tests/test_dbscan.py
@@ -52,8 +52,6 @@ def test_default_cuml_params() -> None:
     cuml_params = get_default_cuml_parameters([CumlDBSCAN], ["handle", "output_type"])
     cuml_params["calc_core_sample_indices"] = False
     spark_params = DBSCAN()._get_cuml_params_default()
-    # TODO: support algorithm parameter added in cuML 24.04
-    cuml_params.pop("algorithm")
     assert cuml_params == spark_params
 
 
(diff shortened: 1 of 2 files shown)


Er1cCheng:
build

Er1cCheng:
build

python/tests/test_dbscan.py
```diff
@@ -166,13 +164,17 @@ def test_dbscan_numeric_type(gpu_number: int, data_type: str) -> None:
     ids=idfn,
 )
 @pytest.mark.parametrize("data_type", cuml_supported_data_types)
-@pytest.mark.parametrize("max_record_batch", [100, 10000])
+@pytest.mark.parametrize(
+    "max_record_batch", [pytest.param(100, marks=pytest.mark.slow), 10000]
+)
+@pytest.mark.parametrize("algorithm", ["brute", "rbc"])
```
eordentlich, line 170: Is this test run time still reasonable?

Er1cCheng: Yes, it's actually pretty fast given it runs on 4 GPUs.

Er1cCheng: And that some parameters are marked as slow so not all of them are executed for CI.

lijinf2 (approved):
Looks good.

PR #623, opened 2024-04-16
work around to support pyspark 3.3 with rapids 24.04

eordentlich:
pyspark 3.3.x uses pandas iteritems method on dfs which has been removed in pandas 2.0.0 .    

This at least allows our 3.3 tests to pass.


lijinf2 (approved):
This looks surprisingly nice. 

Can the same pattern apply to class level? like: 
pyspark.ml.classification.LogisticRegression = spark_rapids_ml.classification.LogisticRegression

PR #642, opened 2024-05-08
Fix NearestNeighbors _ensureIdCol to check whether id_col in df.columns instead of relying on isSet(idCol)

lijinf2:


lijinf2:
build

python/src/spark_rapids_ml/knn.py
```diff
@@ -151,7 +151,7 @@ def _ensureIdCol(self, df: DataFrame) -> DataFrame:
         id_col_name = self.getIdCol()
         df_withid = (
             df
-            if self.isSet("idCol")
+            if id_col_name in df.columns
```
eordentlich, line 154: The following case is still broken it seems.   idCol was not set, but queryDf contains a column with the same name as the default id column.   The above check for this case on self.isSet("idCol") would not work as expected because of the copy.   Seems like relying on isSet leads to very confusing and convoluted logic because of its behavior under copy.   Any alternative?

lijinf2: Revised. The alternative first checks whether getIdCol() is inside the df.
And it throws a ValueError if query_df has a column with the same default name but idCol was not set.

lijinf2:
build

eordentlich:
build

eordentlich (approved):
👍

PR #655, opened 2024-05-13
Support CPUNearestNeighbor for benchmarking exact nearest neighbors. 

lijinf2:
cpu LSH will be moved to bench_approx_nearest_neighbors.py for benchmarking with GPU IVFFlat.

### DIFF
--- a/python/benchmark/benchmark/bench_nearest_neighbors.py
+++ b/python/benchmark/benchmark/bench_nearest_neighbors.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 import time
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from pyspark.ml.feature import VectorAssembler
 from pyspark.ml.functions import array_to_vector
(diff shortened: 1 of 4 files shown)


lijinf2:
build

python/tests/test_benchmark.py
```diff
@@ -0,0 +1,116 @@
...
+
+    with CleanSparkSession({}) as spark:
+
+        def py_func(id: int) -> List[int]:
+            return X[id].tolist()
+
+        from pyspark.sql.functions import udf
+
+        spark_func = udf(py_func, "array<float>")
+        df = spark.range(len(X)).select("id", spark_func("id").alias("features"))
```
eordentlich, line 71: Any advantage to doing this way vs createDataFrame from pandas df?

python/benchmark/benchmark/bench_nearest_neighbors.py
```diff
@@ -65,22 +113,27 @@ def run_once(
...
             input_cols = [c for c in train_df.schema.names]
 
+        query_df = train_df.sample(
+            withReplacement=False, fraction=fraction_sampled_queries, seed=seed
+        )
+
+        def cache_df(dfA: DataFrame, dfB: DataFrame) -> Tuple[DataFrame, DataFrame]:
+            dfA = dfA.cache()
+            dfB = dfB.cache()
+            dfA.count()
```
eordentlich, line 123: Did you verify that count actually caches the dataframe?   I think sometimes it can be short-circuited via metadata (e.g. parquet files).

python/tests/test_benchmark.py
lijinf2: It seems does not throw out a "task size larger than 1000k" warning on large dataset, but it looks no different on small dataset.

python/benchmark/benchmark/bench_nearest_neighbors.py
lijinf2: Revised.

lijinf2:
build

lijinf2:
build

python/benchmark/benchmark/bench_nearest_neighbors.py
```diff
@@ -65,22 +113,33 @@ def run_once(
...
+        def cache_df(dfA: DataFrame, dfB: DataFrame) -> Tuple[DataFrame, DataFrame]:
+            dfA = dfA.cache()
+            dfB = dfB.cache()
+
+            def func_dummy(pdf_iter):  # type: ignore
+                import pandas as pd
+
+                yield pd.DataFrame({"dummy": [1]})
+
+            dfA.mapInPandas(func_dummy, schema="dummy int").count()
```
eordentlich, line 129: Better to avoid python udfs for this kind of thing but probably ok.

eordentlich (approved):
👍

PR #670, opened 2024-06-11
Resolve CI failure by converting X_np to X_pylist and mapping indices to spark ids

lijinf2:
The CI failure appears in pyspark 3.3.4 and python 3.9. 

Seems pyspark.createDataFrame does not recognize numpy array when in 3.3.4. 

Also fix an error related to non-continuous spark ids.

And improve the tolerance of two neighbors having the same distance to a query.

### DIFF
--- a/python/tests/test_benchmark.py
+++ b/python/tests/test_benchmark.py
@@ -33,17 +33,22 @@ def get_sgnn_res(
 
 
 def assert_knn_equal(
-    knn_df: DataFrame, id_col_name: str, distances: np.ndarray, indices: np.ndarray
+    knn_df: DataFrame,
+    id_col_name: str,
+    distances: np.ndarray,
+    indices: np.ndarray,
+    total_tol: float = 1e-3,
 ) -> None:
     res_pd: pd.DataFrame = knn_df.sort(f"query_{id_col_name}").toPandas()
     mg_indices: np.ndarray = np.array(res_pd["indices"].to_list())
     mg_distances: np.ndarray = np.array(res_pd["distances"].to_list())
 
-    assert array_equal(mg_indices, indices)
     assert array_equal(mg_distances, distances)
 
+    # set total_tol because two nearest neighbors may have the same distance to the query
+    assert array_equal(mg_indices, indices, total_tol=total_tol)
 
-@pytest.mark.slow
+
 def test_cpunn_withid() -> None:
 
     n_samples = 1000
@@ -81,7 +86,6 @@ def test_cpunn_withid() -> None:
         assert_knn_equal(knn_df, "id", sg_distances, sg_indices)
 
 
-# @pytest.mark.slow
 def test_cpunn_noid() -> None:
 
     n_samples = 1000
@@ -98,7 +102,9 @@ def test_cpunn_noid() -> None:
 
     with CleanSparkSession({}) as spark:
 
-        df = spark.createDataFrame(X)
+        X_pylist = X.tolist()
+        df = spark.createDataFrame(X_pylist)
+
         from pyspark.sql.functions import array
 
         df = df.select(array(df.columns).alias("features"))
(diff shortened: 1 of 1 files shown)


lijinf2:
build

python/tests/test_benchmark.py
```diff
@@ -33,17 +33,22 @@ def get_sgnn_res(
...
 ) -> None:
     res_pd: pd.DataFrame = knn_df.sort(f"query_{id_col_name}").toPandas()
     mg_indices: np.ndarray = np.array(res_pd["indices"].to_list())
     mg_distances: np.ndarray = np.array(res_pd["distances"].to_list())
 
-    assert array_equal(mg_indices, indices)
     assert array_equal(mg_distances, distances)
 
+    # set total_tol because two nearest neighbors may have the same distance to the query
+    assert array_equal(mg_indices, indices, total_tol=total_tol)
```
eordentlich, line 49: In that case, if indices differ should check that the distances are the same.   This tol could let a bug slip through.

python/tests/test_benchmark.py
```diff
@@ -98,7 +102,9 @@ def test_cpunn_noid() -> None:
 
     with CleanSparkSession({}) as spark:
 
-        df = spark.createDataFrame(X)
```
eordentlich, line 101: Did this ever work?

python/tests/test_benchmark.py
```diff
@@ -110,7 +116,12 @@ def test_cpunn_noid() -> None:
         df_withid, _, knn_df = mg_model.kneighbors(df)
 
         pdf: pd.DataFrame = df_withid.sort(alias.row_number).toPandas()
-        X = np.array(pdf["features"].to_list())
+        X_vec = np.array(pdf["features"].to_list())
+        distances, indices = get_sgnn_res(X_vec, X_vec, n_neighbors)
+
+        X_sparkid = np.array(pdf[alias.row_number].to_list())
+        indices_maped_to_sparkid = X_sparkid[
```
eordentlich, line 123: maped -> mapped

python/tests/test_benchmark.py
```diff
@@ -110,7 +116,12 @@ def test_cpunn_noid() -> None:
...
+            indices
+        ]  # note spark created ids are non-continuous
 
-        distances, indices = get_sgnn_res(X, X, n_neighbors)
-        assert_knn_equal(knn_df, alias.row_number, distances, indices)
+        assert_knn_equal(knn_df, alias.row_number, distances, indices_maped_to_sparkid)
```
eordentlich, line 127: maped -> mapped

eordentlich (approved):

python/tests/test_benchmark.py
```diff
@@ -81,7 +86,6 @@ def py_func(id: int) -> List[int]:
         assert_knn_equal(knn_df, "id", sg_distances, sg_indices)
 
 
-# @pytest.mark.slow
```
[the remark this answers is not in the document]
lijinf2: this was a comment and the test case does not take too long (10s). Let me know if marking it slow is better.

python/tests/test_benchmark.py
lijinf2: Created a follow-up PR to check the returned distances. The checking seemed redundant due to the line "assert array_equal(mg_distances, distances)" which already gets returned distances checked. Perhaps we should recalculate the distance according to indices and then check whether the recalculated distances are equal. Quite troublesome if so but let me know if your comments meant something else.

python/tests/test_benchmark.py
lijinf2: Yes, it works in pyspark 3.4.3

python/tests/test_benchmark.py
eordentlich: Ah ok.   I agree that the returned minimum distances matching is a good check and may be sufficient.   Is there then a need to check the indices at all?   Do we have confidence that the distances are correct for the corresponding indices?

PR #673, opened 2024-06-17
updates for 24.06 rapids compatibility

eordentlich:
mostly version updates +

- changes for some method signature changes for ucx
- change penalty param type change for LR
- default bagging sample rate change
- minor refactor approx knn test to avoid deprecation warning in pytest (tests shouldn't return anything).


eordentlich:
build

lijinf2 (approved):
Looks good!

PR #682, opened 2024-06-25
Support double precision in LogisticRegression

lijinf2:


lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -860,6 +860,8 @@ class LogisticRegression(
         If False, always uses dense array. This is favorable if the majority of VectorUDT vectors are DenseVector.
         If True, always uses sparse array. This is favorable if the majority of the VectorUDT vectors are SparseVector.
         Note this is only supported in spark >= 3.4.
+    float32_inputs: bool (default=True):
```
eordentlich, line 863: We have opted to not add this as a documented constructor argument.   And document only here:  https://nvidia.github.io/spark-rapids-ml/compatibility.html#single-vs-double-precision-inputs  
Should we revisit that?

python/src/spark_rapids_ml/core.py
```diff
@@ -176,13 +176,13 @@ def _get_unwrapped_vec_cols(feature_col: Column) -> List[Column]:
...
+
+    feature_type = FloatType() if float32_input is True else DoubleType()
     return [
         features_unwrapped_vec_col.type.alias(alias.featureVectorType),
         features_unwrapped_vec_col.size.alias(alias.featureVectorSize),
         features_unwrapped_vec_col.indices.alias(alias.featureVectorIndices),
-        # Note: the value field is double array type, cast it to float32 array type
-        # for speedup following repartitioning.
-        features_unwrapped_vec_col.values.cast(ArrayType(FloatType())).alias(
+        features_unwrapped_vec_col.values.cast(ArrayType(feature_type)).alias(
```
eordentlich, line 185: If float32_input = true can we avoid the redundant cast to double?

python/tests/test_logistic_regression.py
```diff
@@ -1986,3 +2021,84 @@ def sparse_to_df(X: csr_matrix, y: List[float]) -> DataFrame:
...
+    fit_intercept: bool,
+    reg_factors: Tuple[float, float],
+    feature_type: str,
+    data_shape: Tuple[int, int],
+    max_record_batch: int,
+    n_classes: int,
+    gpu_number: int,
+) -> None:
+
+    random_bool = random.choice([True, False])
```
eordentlich, line 2045: Would be good to make these deterministic via seed for reproducibility.

python/tests/test_logistic_regression.py
```diff
@@ -998,8 +1008,23 @@ def test_multiclass(
         gpu_number=gpu_number,
         reg_param=0.1,
         tolerance=tolerance,
+        set_float32_inputs=True,
```
eordentlich, line 1011: set_float32_inputs=(data_type == np.float32) or similar would be simpler logic than repeating the whole instantiation.

python/tests/test_logistic_regression.py
```diff
@@ -305,9 +299,16 @@ def test_classifier(
...
     tolerance: float = 0.001,
     convert_to_sparse: bool = False,
+    set_float32_inputs: Optional[bool] = None,
 ) -> LogisticRegression:
     standardization: bool = False
 
+    float32_inputs = set_float32_inputs
+    if float32_inputs is None:
+        float32_inputs = True
+        if data_type == np.float64 and max_record_batch > 1000:
```
eordentlich, line 309: Add a comment explaining reason for this logic.

lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/classification.py
lijinf2: Did not notice that. Revised to be consistent with compatibility.html.

python/src/spark_rapids_ml/core.py
lijinf2: Good catch!

python/tests/test_logistic_regression.py
lijinf2: Done.

python/tests/test_logistic_regression.py
lijinf2: Need to test the case converting float64 to float32. 
Revised to use randomized value.

python/tests/test_logistic_regression.py
lijinf2: Revised the use randomized value.

eordentlich (approved):
👍

PR #691, opened 2024-07-17
Replace withColumn with select for columnar dataframe

lijinf2:
When withColumn is used inside a for loop to add new columns iteratively, it gets slower as the number of columns in the dataframe increases.

### DIFF
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -1286,6 +1286,7 @@ class _CumlModel(Model, _CumlParams, _CumlCommon):
             feature_type: Union[Type[FloatType], Type[DoubleType]] = FloatType
             if not self._float32_inputs and any_double_types:
                 feature_type = DoubleType
+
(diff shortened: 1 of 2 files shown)


lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -1306,6 +1304,12 @@ def _pre_process_data(
                     raise ValueError(
                         "All columns must be integral types or float/double types."
                     )
+
+            taglen = len("_c3BhcmstcmFwaWRzLW1sCg==")
```
eordentlich, line 1308: Better to define the tag in a new variable so all references to it are consistent.

lijinf2:
build

eordentlich (dismissed):
👍

lijinf2:
build

eordentlich (approved):
👍

PR #693, opened 2024-07-22
Fix some test cases of random forest classifier that do not work with multiple GPUs

lijinf2:
So that all nightly test cases get passed when using 2 GPUs.


lijinf2:
build

eordentlich (commented):
Good catch, but on further thought would it be better to change the test so that it works on 2 gpus?   Possible that ci upgrades in the future will allow 2 gpu testing.

lijinf2:
build

eordentlich (approved):
👍

PR #700, opened 2024-08-01
Support int64 index dtype in sparse logistic regression

lijinf2:
Meanwhile update the test environment to cuml 24.08 nightly.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -63,7 +63,7 @@ _default_conf = {
     "spark.python.worker.reuse": "false",
     "spark.driver.host": "127.0.0.1",
     "spark.task.maxFailures": "1",
-    "spark.driver.memory": "5g",
+    "spark.driver.memory": "128g",
     "spark.sql.execution.pyspark.udf.simplifiedTraceback.enabled": "false",
     "spark.sql.pyspark.jvmStacktrace.enabled": "true",
     "spark.sql.execution.arrow.pyspark.enabled": "true",
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -63,7 +63,7 @@ _default_conf = {
     "spark.python.worker.reuse": "false",
     "spark.driver.host": "127.0.0.1",
     "spark.task.maxFailures": "1",
-    "spark.driver.memory": "128g",
+    "spark.driver.memory": "32g",
     "spark.sql.execution.pyspark.udf.simplifiedTraceback.enabled": "false",
     "spark.sql.pyspark.jvmStacktrace.enabled": "true",
     "spark.sql.execution.arrow.pyspark.enabled": "true",
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -59,7 +59,9 @@ def tmp_path() -> Generator[str, None, None]:
 
 
 _default_conf = {
-    "spark.master": f"local[{_gpu_number}]",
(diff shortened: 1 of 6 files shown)


lijinf2:
build

python/tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def tmp_path() -> Generator[str, None, None]:
     "spark.python.worker.reuse": "false",
     "spark.driver.host": "127.0.0.1",
     "spark.task.maxFailures": "1",
-    "spark.driver.memory": "5g",
+    "spark.driver.memory": "128g",
```
eordentlich, line 66: Will this severely limit where this can run?

python/tests/test_logistic_regression.py
```diff
@@ -2092,3 +2099,58 @@ def test_quick_double_precision(
         reg_factors=reg_factors,
         float32_inputs=float32_inputs,
     )
+
+
+@pytest.mark.slow
+def test_sparse_int64() -> None:
+
+    gpu_number = 1
+    data_shape = (int(1e7), 30)
```
eordentlich, line 2108: Will this give only 300 million nonzero?   Thats below the max int limit to be tested?

lijinf2:
build

python/tests/conftest.py
lijinf2: Revised to 32g. The added test leveraged disk and finished in about 10 mins. Seems acceptable as the dataset itself is large. 

Tried setting this to 960g, while physical memory has only 128g. All tests in test_logistic_regression.py run fine. Seems a larger value did not limit the tests to run.

python/tests/test_logistic_regression.py
lijinf2: Revised to use larger dataset and SparseDataGen.

lijinf2:
build

python/tests/test_logistic_regression.py
```diff
@@ -2092,3 +2108,110 @@ def test_quick_double_precision(
...
+    data_gen = SparseRegressionDataGen(data_gen_args)
+    df, _, _ = data_gen.gen_dataframe_and_meta(conftest._spark)
+    df = df.cache()
+
+    # check nnz > int32.max
+    def get_nnz(sparse_vec: SparseVector) -> int:
+        return len(sparse_vec.values)
+
+    udf_get_nnz = udf(get_nnz, LongType())
+    nnz_df = df.select(udf_get_nnz("features").alias("nnz"))
```
eordentlich, line 2190: Is this slow?  udf is invoked on each row I believe.

python/tests/test_logistic_regression.py
```diff
@@ -1,10 +1,20 @@
+import math
+import os
+import sys
+
+file_path = os.path.abspath(__file__)
```
eordentlich, line 5: Would it be better/cleaner to set PYTHONPATH env var in run_test.sh .

python/tests/test_logistic_regression.py
lijinf2: Good catch. I replaced it with mapInPandas. Runtime was reduced by 45 seconds.

python/tests/test_logistic_regression.py
lijinf2: Yeah, revised to use PYTHONPATH env variable.

lijinf2:
build

eordentlich (approved):
👍

PR #703, opened 2024-08-08
Support IVFPQ algorithm in ApproximateNearestNeighbors

lijinf2:


lijinf2:
build

python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -95,45 +95,136 @@ def test_example(
...
+            n_neighbors=self.n_neighbors,
+            metric=self.metric,
+        )
+        cuml_ivfflat.fit(self.X)
+        cumlsg_distances, cumlsg_indices = cuml_ivfflat.kneighbors(self.X)
+        if self.metric == "euclidean" or self.metric == "l2":
+            cumlsg_distances **= 2  # square up cuml distances to get l2 distances
+
+        # compare cuml sg with given results
+        avg_recall_cumlann = self.cal_avg_recall(cumlsg_indices)
```
eordentlich, line 169: Would it be better to instead compare the given indices and distances  directly to what is returned by sg (instead of comparing the respective recalls and diff from the exact case)?   Or is the distributed setting too different from sg for this?

python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -312,6 +453,6 @@ def test_ivfflat_wide_matrix(
     import time
 
     start = time.time()
-    test_ivfflat(combo=combo, data_shape=data_shape, data_type=data_type)
+    test_ann_algorithm(combo=combo, data_shape=data_shape, data_type=data_type)
```
eordentlich, line 456: I'm not seeing the algorithm field for the combo params for this test.  How could this pass?

lijinf2: This does not test ivfpq. It tests the performance of multi_cols input format, and is part of a previous PR: https://github.com/NVIDIA/spark-rapids-ml/pull/691

eordentlich: But it now is changed to go via test_ann_algorithm which seems to expect combo[0] to be the algorithm

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Yeah, distributed setting sometimes can be too different from sg. distance gap can be as large as 0.05 (current tolerance is 0.001). 

In fact, same error happens when running SG twice, the distance gap can be large. 

It seems the returned distances are sensitive to the quantization codebook. And codebooks are not the same when running SG multiple times.

lijinf2:
build

lijinf2:
build

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Good finding! This is a slow test so was not run on pre-merge ci. Revised.

eordentlich (approved):
👍

PR #715, opened 2024-08-22
Increase tolerance for testing ivfpq due to quantization and randomness

lijinf2:
The test case occasionally fails, at asserting the returned distances are equal when running kneighbors and approxSimilarityJoin. 
In fact, this may not be the case for ivfpq because returned distances seem sensitive to codebook and randomness.

### DIFF
--- a/python/tests/test_approximate_nearest_neighbors.py
+++ b/python/tests/test_approximate_nearest_neighbors.py
@@ -332,7 +332,7 @@ def test_ann_algorithm(
             if algorithm == "ivfpq":
                 # returned distances can be slightly different when running ivfpq multiple times due to quantization and randomness
                 assert array_equal(
-                    r1_distances, r2_distances, unit_tol=tolerance, total_tol=tolerance
+                    r1_distances, r2_distances, unit_tol=tolerance, total_tol=1e-3
                 )
             else:
                 assert array_equal(r1_distances, r2_distances, tolerance)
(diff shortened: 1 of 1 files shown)


python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -329,7 +329,13 @@ def assert_row_equal(r1: Row, r2: Row) -> None:
             assert r1[f"query_{id_col}"] == r2[f"query_{id_col}"]
             r1_distances = r1["distances"]
             r2_distances = r2["distances"]
-            assert array_equal(r1_distances, r2_distances, tolerance)
+            if algorithm == "ivfpq":
+                # returned distances can be slightly different when running ivfpq multiple times due to quantization and randomness
+                assert array_equal(
+                    r1_distances, r2_distances, unit_tol=tolerance, total_tol=tolerance
```
eordentlich, line 335: total tolerance > 0 means that some entries can be way off.   Is this needed vs just increasing unit_tol?  What would cause some entries to have huge error?

lijinf2: Removed the total tolerance, the test sometimes fails, at the same place. 
Set the total tolerance to e-3 and repeat the test five times, all get passed. 
Set the total tolerance to e-4, the test fails at the same place. 

The total tolerance is needed. Returned distances are different even when running the SG class twice. Codebook difference maybe be the cause.

lijinf2:
build

eordentlich (approved):
👍 
Let's look into more deterministic runs, but ok in the mean time.

PR #716, opened 2024-08-23
changes to support db 13.3+

eordentlich:
Also adds db version as option to db benchmark script and a gpu only option (i.e. no spark rapids plugin).

Plus some other misc. updates (e.g. to logistic reg. notebooks, spark-rapids versions).


eordentlich:
build

python/benchmark/benchmark/bench_kmeans.py
```diff
@@ -185,11 +187,22 @@ def gpu_cache_df(df: DataFrame) -> DataFrame:
...
 
+            # temporary patch for DB with spark-rapids plugin
+            # this part is not timed so overhead is not critical, but should be reverted
+            # once https://github.com/NVIDIA/spark-rapids/issues/10770 is fixed
+            db_version = os.environ.get("DATABRICKS_RUNTIME_VERSION")
+            if db_version:
+                dim = len(cluster_centers[0])
+                # inject unsupported expr (slice) that is essentially a noop
+                df_for_scoring = df_for_scoring.select(
+                    F.slice(feature_col, 1, dim).alias(feature_col), output_col
```
lijinf2, line 203: Interesting. Any intuition as to why slice can get the hanging resolved?

python/benchmark/databricks/README.md
```diff
@@ -56,6 +58,6 @@ This directory contains shell scripts for running larger scale benchmarks on Dat
   databricks jobs cancel-run <runid> --profile $DB_PROFILE
   ```
 
-5. The created clusters are configured to terminate after 30 min, but can be manually terminated or deleted via the Databricks UI.
+1. The created clusters are configured to terminate after 30 min, but can be manually terminated or deleted via the Databricks UI.
 
-6. Monitor progress periodically in case of a possible hang, to avoid incurring cloud costs in such cases.
+2. Monitor progress periodically in case of a possible hang, to avoid incurring cloud costs in such cases.
```
lijinf2, line 63: Do we keep item number 5, 6 or replace with 1, 2?

python/benchmark/databricks/gpu_etl_cluster_spec.sh
```diff
@@ -0,0 +1,65 @@
...
+        "spark.sql.execution.sortBeforeRepartition": "false",
+        "spark.rapids.sql.python.gpu.enabled": "true",
+        "spark.rapids.memory.pinnedPool.size": "2G",
+        "spark.python.daemon.module": "rapids.daemon_databricks",
+        "spark.rapids.sql.batchSizeBytes": "512m",
+        "spark.sql.adaptive.enabled": "false",
+        "spark.sql.execution.arrow.pyspark.enabled": "true",
+        "spark.sql.files.maxPartitionBytes": "2000000000000",
+        "spark.databricks.delta.optimizeWrite.enabled": "false",
+        "spark.rapids.sql.concurrentGpuTasks": "2"
```
lijinf2, line 32: While we keep this here, do we need to remove it from gpu_cluster_spec.sh (the one without GPU ETL)?

lijinf2 (commented):
Looks good. Minor comments.

python/benchmark/benchmark/bench_kmeans.py
eordentlich: Yes.  It is an unsupported expression in spark-rapids so falls back to cpu and injects columnartorow and rowtocolumnar transformations which do some batching that doesn't (but should after patching) happen otherwise.

python/benchmark/databricks/README.md
eordentlich: Interesting.  Didn't intend to change the numbers, will revert.

python/benchmark/databricks/gpu_etl_cluster_spec.sh
eordentlich: Yes.  Will delete there.  Good catch.

eordentlich:
build

lijinf2 (approved):

PR #723, opened 2024-08-29
avoid testing ivfpq approxSimilarityJoin due to non-deterministic results when compared with kneighbors

lijinf2:


lijinf2:
build

eordentlich (approved):
👍

PR #725, opened 2024-08-30
initial attempt at enabling no package import change

eordentlich:
still needs documentation, test, lint type checking etc.   The file `python/test_no_import_change.py` can be tried for some basic testing.

```bash
# to run on cpu
CUDA_VISIBLE_DEVICES=0 python python/test_no_import_change.py 0.2
# run on gpu 
CUDA_VISIBLE_DEVICES=0 python -m spark_rapids_ml python/test_no_import_change.py 0.2
# gpu using spark-submit (for cpu just omit the __main__.py)
CUDA_VISIBLE_DEVICES=0 spark-rapids-submit --master local[2] \
python/test_no_import_change.py 0.2
```


python/src/spark_rapids_ml/install.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _getattr(attr: str) -> Any:
+        frame = sys._getframe()
+        assert frame.f_back
+        calling_path = frame.f_back.f_code.co_filename
+        if any(
+            [
+                (
+                    f"pyspark/ml/{mod_name}" in calling_path
+                    or f"spark_rapids_ml/{mod_name}" in calling_path
```
lijinf2, line 52: Would this always be true? The calling_path must come from either pyspark or spark_rapids_ml I think.
If so, the if condition will always be true.

python/test_no_import_change.py
```diff
@@ -0,0 +1,91 @@
...
+print(f"cvModel.getNumFolds(): {cvModel.getNumFolds()}")
+print(f"cvModel.avgMetrics[0]: {cvModel.avgMetrics[0]}")
+print(f"evaluator.evaluate(cvModel.transform(dataset)): {evaluator.evaluate(cvModel.transform(dataset))}")
+path = tempfile.mkdtemp()
+model_path = path + "/model"
+cvModel.write().save(model_path)
+cvModelRead = CrossValidatorModel.read().load(model_path)
+print(f"cvModelRead.avgMetrics: {cvModelRead.avgMetrics}")
+print(f"evaluator.evaluate(cvModel.transform(dataset)): {evaluator.evaluate(cvModel.transform(dataset))}")
+print(f"evaluator.evaluate(cvModelRead.transform(dataset)): {evaluator.evaluate(cvModelRead.transform(dataset))}")
```
lijinf2, line 91: Is it possible to check whether both the estimator and the model are from spark_rapids_ml (or both come from pyspark.ml)?

lijinf2 (commented):
The idea looks good! Some questions come to my mind.

python/src/spark_rapids_ml/install.py
eordentlich: No it is not always true.   the condition will only be true if the import statement or module attribute access to reach here happened in the specified modules (our modules or select pyspark/ml modules).  In these cases we want to use the pyspark.ml version of the class always.   This allows us to subclass pyspark ml classes like in our CrossValidator and also use pyspark classes to validate parameters.

eordentlich:
build

eordentlich:
build

wbo4958:
I just executed the `python -m spark_rapids_ml` but showed nothing. Is this expected? or should print the help usage?

eordentlich:
Yes we could print usage in that case.  Good suggestion.

python/src/spark_rapids_ml/install.py
```diff
@@ -0,0 +1,80 @@
...
+                )
+                for mod_name in _accelerated_attributes.keys()
+            ]
+        ) or (attr not in _accelerated_attributes[mod_name]):
+            try:
+                attr_val = getattr(_pyspark_modules[mod_name], attr)
+            except:
+                raise AttributeError("No attribute '{attr}'")
```
wbo4958, line 69: `f"No attribute '{attr}'"`

python/tests_no_import_change/test_no_import_change.py
```diff
@@ -0,0 +1,145 @@
...
+schema = "features array<float>, label float"
+
+spark = SparkSession.builder.getOrCreate()
+df = spark.createDataFrame(data, schema=schema)
+df = df.select(array_to_vector(col("features")).alias("features"), "label")
+df.show()
+
+lr_estimator = LogisticRegression()
+print(lr_estimator.setFeaturesCol("features"))
+
```
wbo4958, line 55: I added `lr_estimator.setStandardization(False)` but it ran with error

`AttributeError: 'LogisticRegression' object has no attribute 'setStandardization'. Did you mean: 'getStandardization'?`

No idea why.

eordentlich: Turns out we never defined setStandardization for spark_rapids_ml LogisticRegression. adding to this PR.

eordentlich:
build

eordentlich:
build

lijinf2 (approved):

python/tests_no_import_change/test_no_import_change.py
wbo4958: Yeah, In order to make no code change, seems we need to make sure that spark-rapids-ml contains the APIs of spark.ml

wbo4958: but it could be in a follow up.

eordentlich: Yes.  Current goal was to eliminate package import change.   We have quite a few gaps to address for full api compatibility.

PR #738, opened 2024-09-16
better match spark mllib in the case of non-occurrence of labels

eordentlich:
in crossvalidation it is possible for for some validation folds to not have all label values occur.   This PR fixes a dividebyzero error that can occur in these cases in our code that does not occur in Spark mllib.

Should resolve a recent intermittent failures in our tests.


eordentlich:
build

lijinf2 (approved):
Looks good!

PR #739, opened 2024-09-17
add explicit test for fewer classes in test vs train

eordentlich:
better to have an explicit test for the case addressed by https://github.com/NVIDIA/spark-rapids-ml/pull/738


eordentlich:
build

lijinf2 (approved):
Looks good!

Issue #740, opened 2024-09-18
[Bug] test_umap.py::test_spark_umap FAILED


NvTimLiu:
Detailed failure logs as below:

```

 =================================== FAILURES ===================================
 ___________ test_spark_umap[vector-float32-10-iris-False-0.55-100-2] ___________
 
 n_parts = 2, gpu_number = 1, n_rows = 100, sampling_ratio = 0.55
 supervised = False, dataset = 'iris', n_neighbors = 10
 dtype = <class 'numpy.float32'>, feature_type = 'vector'
 
     @pytest.mark.parametrize("n_parts", [2, 9])
     @pytest.mark.parametrize("n_rows", [100, 500])
     @pytest.mark.parametrize("sampling_ratio", [0.55, 0.9])
     @pytest.mark.parametrize("supervised", [True, False])
     @pytest.mark.parametrize("dataset", ["digits", "iris"])
     @pytest.mark.parametrize("n_neighbors", [10])
     @pytest.mark.parametrize("dtype", cuml_supported_data_types)
     @pytest.mark.parametrize("feature_type", pyspark_supported_feature_types)
     @pytest.mark.slow
     def test_spark_umap(
         n_parts: int,
         gpu_number: int,
         n_rows: int,
         sampling_ratio: float,
         supervised: bool,
         dataset: str,
         n_neighbors: int,
         dtype: np.dtype,
         feature_type: str,
     ) -> None:
         result = _run_spark_test(
             n_parts,
             gpu_number,
             n_rows,
             sampling_ratio,
             supervised,
             dataset,
             n_neighbors,
             dtype,
             feature_type,
         )
     
         if not result:
             result = _run_spark_test(
                 n_parts,
                 gpu_number,
                 n_rows,
                 sampling_ratio,
                 supervised,
                 dataset,
                 n_neighbors,
                 dtype,
                 feature_type,
             )
     
 >       assert result
 E       assert False
 
 tests/test_umap.py:194: AssertionError
 ----------------------------- Captured stdout call -----------------------------
 Local UMAP trustworthiness score : 0.91
 Spark UMAP trustworthiness score : 0.75
 Local UMAP trustworthiness score : 0.91
 Spark UMAP trustworthiness score : 0.76
 ----------------------------- Captured stderr call -----------------------------
  - spark_rapids_ml.umap.UMAP - INFO - Initializing cuml context
  - spark_rapids_ml.umap.UMAP - INFO - Loading data into python worker memory
[...]

PR #745, opened 2024-09-27
Fix UMAP test failure

rishic3:
Remove subsampling for Spark UMAP so that both cuML / Spark implementations get full dataset for consistency. 
Will also resolve [failing test case](https://github.com/NVIDIA/spark-rapids-ml/issues/740).

### DIFF
--- a/python/tests/test_umap.py
+++ b/python/tests/test_umap.py
@@ -369,9 +369,11 @@ def test_umap_broadcast_chunks(gpu_number: int, BROADCAST_LIMIT: int) -> None:
 def test_umap_sample_fraction(gpu_number: int) -> None:
     from cuml.datasets import make_blobs
 
+    n_rows = 5000
+
     X, _ = make_blobs(
-        5000,
-        3000,
+        n_rows,
+        10,
         centers=42,
         cluster_std=0.1,
         dtype=np.float32,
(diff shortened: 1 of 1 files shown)


rishic3:
build

python/tests/test_umap.py
```diff
@@ -102,7 +101,7 @@ def _spark_umap_trustworthiness(
             )
 
         data_df = data_df.repartition(n_parts)
-        umap_estimator.setFeaturesCol(features_col).setSampleFraction(sampling_ratio)
```
eordentlich, line 105: Is there still some test of the sample fraction param?

eordentlich:
build

python/tests/test_umap.py
rishic3: Added test

python/tests/test_umap.py
```diff
@@ -375,3 +364,47 @@ def assert_umap_model(model: UMAPModel) -> None:
...
+    with CleanSparkSession() as spark:
+        pyspark_type = "float"
+        feature_cols = [f"c{i}" for i in range(X.shape[1])]
+        schema = [f"{c} {pyspark_type}" for c in feature_cols]
+        df = spark.createDataFrame(X.tolist(), ",".join(schema))
+        df = df.withColumn("features", array(*feature_cols)).drop(*feature_cols)
+
+        sample_fraction = 0.5
+        umap = (
+            UMAP(num_workers=gpu_number)
```
eordentlich, line 390: Can random_state be set to make it deterministic?  Otherwise, seems will have same failure rate as current test.

rishic3: thanks for catching that. fixed

eordentlich:
build

eordentlich (approved):
👍

PR #746, opened 2024-09-28
update package version tags to 24.10, switch to nightly 24.10 rapids

eordentlich:
…in ci and db benchmark, update spark-rapids to 24.08.1

note: rapids no longer upper bounding numpy, so adding that as pyspark <= 3.5 not compatible with numpy 2.0

### DIFF
--- a/python/benchmark/databricks/init-pip-cuda-11.8.sh
+++ b/python/benchmark/databricks/init-pip-cuda-11.8.sh
@@ -24,17 +24,12 @@ ln -s /usr/local/cuda-11.8 /usr/local/cuda
 
 # install cudf and cuml
 # using ~= pulls in micro version patches
-# /databricks/python/bin/pip install cudf-cu11~=${RAPIDS_VERSION} \
-#     cuml-cu11~=${RAPIDS_VERSION} \
-#     cuvs-cu11~=${RAPIDS_VERSION} \
-#     pylibraft-cu11~=${RAPIDS_VERSION} \
-#     rmm-cu11~=${RAPIDS_VERSION} \
-#     --extra-index-url=https://pypi.nvidia.com
-
-/databricks/python/bin/pip install \
-    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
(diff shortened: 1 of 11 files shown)


python/benchmark/databricks/init-pip-cuda-11.8.sh
```diff
@@ -24,12 +24,17 @@ ln -s /usr/local/cuda-11.8 /usr/local/cuda
...
+#     cuml-cu11~=${RAPIDS_VERSION} \
+#     cuvs-cu11~=${RAPIDS_VERSION} \
+#     pylibraft-cu11~=${RAPIDS_VERSION} \
+#     rmm-cu11~=${RAPIDS_VERSION} \
+#     --extra-index-url=https://pypi.nvidia.com
+
+/databricks/python/bin/pip install \
+    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
+    "cudf-cu11>=24.10.0a0,<=24.10" "dask-cudf-cu11>=24.10.0a0,<=24.10" \
+    "cuml-cu11>=24.10.0a0,<=24.10" "dask-cuda>=24.10.0a0,<=24.10"
```
lijinf2, line 37: Would cuvs be automatically installed even it is not specified in the pip install command?
Seems we introduce cudf, dask-cudf, and dask-cuda to the pip install command. Any reason for that?

lijinf2 (commented):
Overall looks good. Some questions only.

eordentlich:
build

lijinf2:
The latest ci failure seems related to ivfpq unstable codebooks, which leads to unstable recall. We may increase n_lists, M, n_bits, and n_probe to make the recall stable.

eordentlich:
Is this new with 24.10?

lijinf2:
I think so. The test case gets passed with 24.08. It seems the ivf_pq implementation is updated in 24.10. 
BTW, the following algoParams get passed with 24.10 in my local environment: 
{ "nlist": 100, "nprobe": 20, "M": 20, "n_bits": 8,}

eordentlich:
build

lijinf2 (approved):

Issue #749, opened 2024-10-04
[Bug] Test FAILED tests/test_umap.py::test_umap_sample_fraction


yinqingh:
Observed test failure in spark-rapids-ml_nightly#510. Test failure details: 
```
[2024-10-04T06:08:10.749Z] =================================== FAILURES ===================================
[2024-10-04T06:08:10.749Z] __________________________ test_umap_sample_fraction ___________________________
[2024-10-04T06:08:10.749Z] 
[2024-10-04T06:08:10.749Z] gpu_number = 1
[2024-10-04T06:08:10.749Z] 
[2024-10-04T06:08:10.749Z]     def test_umap_sample_fraction(gpu_number: int) -> None:
[2024-10-04T06:08:10.749Z]         from cuml.datasets import make_blobs
[2024-10-04T06:08:10.749Z]     
[2024-10-04T06:08:10.749Z]         n_rows = 5000
[2024-10-04T06:08:10.749Z]         sample_fraction = 0.5
[2024-10-04T06:08:10.749Z]     
[2024-10-04T06:08:10.749Z]         X, _ = make_blobs(
[2024-10-04T06:08:10.749Z]             n_rows,
[2024-10-04T06:08:10.749Z]             10,
[2024-10-04T06:08:10.749Z]             centers=42,
[2024-10-04T06:08:10.749Z]             cluster_std=0.1,
[2024-10-04T06:08:10.749Z]             dtype=np.float32,
[2024-10-04T06:08:10.749Z]             random_state=10,
[2024-10-04T06:08:10.749Z]         )
[2024-10-04T06:08:10.749Z]     
[2024-10-04T06:08:10.749Z]         with CleanSparkSession() as spark:
[2024-10-04T06:08:10.749Z]             pyspark_type = "float"
[2024-10-04T06:08:10.749Z]             feature_cols = [f"c{i}" for i in range(X.shape[1])]
[2024-10-04T06:08:10.749Z]             schema = [f"{c} {pyspark_type}" for c in feature_cols]
[2024-10-04T06:08:10.749Z]             df = spark.createDataFrame(X.tolist(), ",".join(schema))
[2024-10-04T06:08:10.749Z]             df = df.withColumn("features", array(*feature_cols)).drop(*feature_cols)
[2024-10-04T06:08:10.749Z]     
[2024-10-04T06:08:10.749Z]             umap = (
[...]

PR #751, opened 2024-10-08
Fix cuml parameter setting issues in UMAP/DBSCAN

rishic3:
Fixed an issue where cuml parameters in UMAP were not being set due to a silent failure in `_set_param`, where these parameters could not be found in the `_param_mapping`. Mapping these params to themselves fixes this issue, as it is [done in KNN](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-24.10/python/src/spark_rapids_ml/knn.py#L790). 

DBSCAN had the same issue where cuml parameters were never set, but got away with this by [using the Spark API to retrieve parameters](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-24.10/python/src/spark_rapids_ml/clustering.py#L970) and avoiding the use of `cuml_params` dict entirely. 

## Changes:
- Updated both DBSCAN and UMAP to include identity mappings for cuml parameters in `_param_mapping`.
- Updated DBSCAN to use the `cuml_params` dictionary for initialization to be consistent with the other algorithms.
- Updated the `_param_mapping` description to avoid future confusion for cuml algorithms without a PySpark equivalent.
- Added tests for DBSCAN and UMAP to check that cuml parameters are properly set using both the Spark param API and the `cuml_params` attribute, and test under non-default cuml parameter settings. 

This PR also resolves Issue #749, as `random_state` is now properly set, making dataframe sampling deterministic.

### DIFF
--- a/python/src/spark_rapids_ml/clustering.py
+++ b/python/src/spark_rapids_ml/clustering.py
@@ -836,7 +836,7 @@ class DBSCAN(DBSCANClass, _CumlEstimator, _DBSCANCumlParams):
 
         model._num_workers = self.num_workers
         self._copyValues(model)
-        self._copy_cuml_params(model)
+        self._copy_cuml_params(model)  # type: ignore
 
         return model
 
--- a/python/src/spark_rapids_ml/clustering.py
+++ b/python/src/spark_rapids_ml/clustering.py
@@ -502,7 +502,14 @@ class KMeansModel(KMeansClass, _CumlModelWithPredictionCol, _KMeansCumlParams):
 class DBSCANClass(_CumlClass):
     @classmethod
     def _param_mapping(cls) -> Dict[str, Optional[str]]:
-        return {}
+        return {
+            "eps": "eps",
+            "min_samples": "min_samples",
+            "metric": "metric",
+            "algorithm": "algorithm",
+            "max_mbytes_per_batch": "max_mbytes_per_batch",
+            "calc_core_sample_indices": "calc_core_sample_indices",
+        }
 
     def _get_cuml_params_default(self) -> Dict[str, Any]:
         return {
@@ -829,6 +836,7 @@ class DBSCAN(DBSCANClass, _CumlEstimator, _DBSCANCumlParams):
 
         model._num_workers = self.num_workers
         self._copyValues(model)
+        self._copy_cuml_params(model)
 
         return model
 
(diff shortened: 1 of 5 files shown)


python/src/spark_rapids_ml/params.py
```diff
@@ -145,6 +145,9 @@ def _param_mapping(cls) -> Dict[str, Optional[str]]:
         - empty string, if a defined Spark Param should just be silently ignored, or
         - None, if a defined Spark Param should raise an error.
 
+        If the class has no Spark ML equivalent, the cuML parameter names should be mapped to themselves,
+        i.e., `{"cuml_param": "cuml_param"}`.
```
lijinf2, line 149: Seems we do not need this. Existing code directly sets cuml param if the param name matches, as shown here: https://github.com/NVIDIA/spark-rapids-ml/blob/f0fa39cde77e8c6b4b8f598e6740cc1b7579ec06/python/src/spark_rapids_ml/params.py#L342

python/src/spark_rapids_ml/umap.py
```diff
@@ -90,7 +90,25 @@
 class UMAPClass(_CumlClass):
     @classmethod
     def _param_mapping(cls) -> Dict[str, Optional[str]]:
-        return {}
+        return {
+            "n_neighbors": "n_neighbors",
```
lijinf2, line 94: Would there be any problem if we remove such pairs where key==value?

python/src/spark_rapids_ml/clustering.py
```diff
@@ -502,7 +502,14 @@ def _transform_internal(
 class DBSCANClass(_CumlClass):
     @classmethod
     def _param_mapping(cls) -> Dict[str, Optional[str]]:
-        return {}
+        return {
+            "eps": "eps",
+            "min_samples": "min_samples",
+            "metric": "metric",
+            "algorithm": "algorithm",
```
lijinf2, line 509: Similarly, we can try to remove these pairs if key == value.

lijinf2 (commented):
Thanks for the PR! Looks pretty good. 
Just one minor comment.

python/tests/test_umap.py
```diff
@@ -385,15 +394,16 @@ def test_umap_sample_fraction(gpu_number: int) -> None:
         pyspark_type = "float"
         feature_cols = [f"c{i}" for i in range(X.shape[1])]
         schema = [f"{c} {pyspark_type}" for c in feature_cols]
-        df = spark.createDataFrame(X.tolist(), ",".join(schema))
+        df = spark.createDataFrame(X.tolist(), ",".join(schema)).coalesce(1)
```
lijinf2, line 397: Nit: do we need coalesce here or it is optional?

lijinf2:
build

python/tests/test_umap.py
rishic3: Mainly a peace of mind thing to ensure determinism, such that the sampling is applied to a single partition; though it would only really be needed if there were upstream joins/shuffles that could introduce [non-deterministic behavior](https://medium.com/udemy-engineering/pyspark-under-the-hood-randomsplit-and-sample-inconsistencies-examined-7c6ec62644bc).

python/src/spark_rapids_ml/params.py
rishic3: Resolving after our discussion online.

lijinf2 (approved):

eordentlich:
@lijinf2 should @rishic3 add similar tests to this PR for other algo params?

PR #757, opened 2024-10-14
Revise to cuvs ivf_pq, add cosine for ivf_pq and support long_max indices for all ann algorithms.

lijinf2:


lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/knn.py
```diff
@@ -1531,19 +1556,13 @@ def _transform_internal(
...
-                    cuml_alg_params["metric"] == "euclidean"
-                    or cuml_alg_params["metric"] == "l2"
-                ):
-                    distances = distances * distances
-
             if isinstance(distances, cp.ndarray):
                 distances = distances.get()
 
+            # in case a query did not probe any items, indices are filled with int64 max and distances are filled with inf
+            item_row_number = np.append(item_row_number, np.iinfo("int64").max)
```
eordentlich, line 1563: Nice trick!

python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -625,10 +636,14 @@ def test_ivfpq(
 ) -> None:
     """
     (1) Currently the usePrecomputedTables is not used in cuml C++.
-    (2) ivfpq has become unstable in 24.10. It does not get passed with algoParam {"nlist" : 10, "nprobe" : 2, "M": 2, "n_bits": 4} in ci where test_ivfflat is run beforehand. avg_recall shows large variance, depending on the quantization accuracy. This can be fixed by increasing nlist, nprobe, M, and n_bits.
+
+    (2) ivfpq has become unstable in 24.10. It does not get passed with algoParam {"nlist" : 10, "nprobe" : 2, "M": 2, "n_bits": 4} in ci where test_ivfflat is run beforehand. avg_recall shows large variance, depending on the quantization accuracy. This can be fixed by increasing nlist, nprobe, M, and n_bits. Note ivf_pq is non-deterministic, and it seems due to kmeans initialization leveraging runtime values of GPU memory.
+
+    (3) If M is is too small (e.g. 2), the returned distances can be very different from the ground distances.
+    Spark rapids ml may give lower recall than cuvs sg because it aggregates local topk candidates by the returned distances.
```
eordentlich, line 643: I thought cuml has a 'refine distances' api that can be used to fill in the exact distances.

python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -793,3 +808,103 @@ def test_ivfflat_wide_matrix(
...
+)
+@pytest.mark.parametrize("data_type", [np.float32])
+def test_return_fewer_k(
+    algorithm: str,
+    feature_type: str,
+    data_type: np.dtype,
+) -> None:
+    """
+    This tests the corner case where there are less than k neighbors found due to nprobe too small.
+    (1) In ivf_flat and ivf_pq:
```
eordentlich, line 834: Should this behavior be documented in knn api doc string?

lijinf2:
build

lijinf2:
build

python/src/spark_rapids_ml/knn.py
```diff
@@ -1419,23 +1443,20 @@ def _get_cuml_transform_func(
                 metric=self.cuml_params["metric"],
                 topk=cuml_alg_params["n_neighbors"],
             )
+        else:
+            assert cuml_alg_params["algorithm"] in {"ivfpq", "ivf_pq"}
```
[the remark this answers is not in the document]
lijinf2: Good catch. Revised.

python/tests/utils.py
```diff
@@ -92,10 +93,17 @@ def create_pyspark_dataframe(
...
 
     if label is not None:
+        label_dtype = dtype if label_dtype is None else label_dtype
+        label = label.astype(label_dtype)
+        label_pyspark_type = dtype_to_pyspark_type(label_dtype)
+
         label_col = "label_col"
-        schema.append(f"{label_col} {pyspark_type}")
+        schema.append(f"{label_col} {label_pyspark_type}")
+        data_pytype = [
```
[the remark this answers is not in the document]
lijinf2: revised to use pandas.

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Moved it to the class doc string when explaining parameter k.

lijinf2:
build

lijinf2:
build

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Yeah, that seems to be working for SG only. Do we want to support that or use that in our test case?

eordentlich: But we invoke SG classes locally, so can we do the refine in each worker?  Seems important to accurately pick the top k from the partial results.

lijinf2:
build

lijinf2:
build

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Yeah, should be able to call the refine in mapInPandas and after cuvs PQ. A few questions:

(Q1) It seems the refine API returns the exact distances, together with the refined topk indices. The API does not return the approximate (PQ) distances to the refined topk indices. That means, our "ivf_pq" algorithm will be returning exact distances, which is different from the behavior of cuvs "ivf_pq". 

(Q2) Is the "ivf_pq + refine" equivalent to the ivf_pq build_algo of cagra? We have already supported the ivf_pq option of cagra.

lijinf2:
build

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Added the refine API for ivf_pq to return exact distances.

rishic3:
For the Cagra "compression" param, is there a way to create and pass a CompressionParams object in Python?

python/src/spark_rapids_ml/knn.py
```diff
@@ -1310,6 +1316,30 @@ def _cal_cuvs_ivf_flat_params_and_check(
 
         return (ivfflat_index_params, ivfflat_search_params)
 
+    @classmethod
+    def _cal_cuvs_ivf_pq_params_and_check(
```
rishic3, line 1320: nit: should these be _call_ (here and throughout)?

python/src/spark_rapids_ml/knn.py
```diff
@@ -1397,12 +1427,17 @@ def _get_cuml_transform_func(
...
-            check_fn = (
-                self._cal_cagra_params_and_check
-                if cuml_alg_params["algorithm"] == "cagra"
-                else self._cal_cuvs_ivf_flat_params_and_check
-            )
+        if (
+            cuml_alg_params["algorithm"] != "brute"
+        ):  # brute links to CPUNearestNeighborsModel of benchmark.bench_nearest_neighbors
+            if cuml_alg_params["algorithm"] == "cagra":
+                check_fn = self._cal_cagra_params_and_check
```
rishic3, line 1434: Would it be helpful to check if metric == "sqeuclidean" at initialization if given a cagra algo, rather than waiting till transform to catch - or just set it by default if given cagra?

python/tests/utils.py
```diff
@@ -113,6 +135,9 @@ def create_pyspark_dataframe(
             .drop(*feature_cols)
         )
         feature_cols = "features"
+    else:
+        # When df has many columns (e.g. 3000), the select here breaks the runtime slowdown observed at calling df.withColumn.
```
eordentlich, line 139: breaks or fixes?

python/tests/test_approximate_nearest_neighbors.py
```diff
@@ -808,9 +839,110 @@ def test_ivfflat_wide_matrix(
     data_shape: Tuple[int, int],
     data_type: np.dtype,
 ) -> None:
+    """
+    It seems adding a column with df.withColumn can be very slow, if df already has many columns (e.g. 3000).
```
eordentlich, line 843: This comment would make more sense closer to the actual withColumn .

python/src/spark_rapids_ml/knn.py
```diff
@@ -1504,35 +1526,57 @@ def _transform_internal(
...
 
-            # Note cuML kneighbors applys an extra square root on the l2 distances.
-            # Here applies square to obtain the actual l2 distances.
-            if isinstance(nn_object, cumlSGNN):
-                if (
-                    cuml_alg_params["metric"] == "euclidean"
-                    or cuml_alg_params["metric"] == "l2"
-                ):
-                    distances = distances * distances
+                # in case refine API reset inf distances to 0.
```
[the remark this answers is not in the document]
lijinf2: Yes, refine changes the indices.

python/src/spark_rapids_ml/knn.py
lijinf2: the checking seems meaningful only when both metric and algorithm have been set. This may always be the case in initialization. For example, the application code could set metric in __init__ but set algorithm through setter.

lijinf2:
> For the Cagra "compression" param, is there a way to create and pass a CompressionParams object in Python?

Yeah, it should work as long as a compression object is created and passed into algoParams as {"compression":compression_obj}.
Did not add that to the PR. Seems it is not yet listed as one of the cuvs python APIs with example code.

python/src/spark_rapids_ml/knn.py
```diff
@@ -1504,35 +1526,57 @@ def _transform_internal(
...
+
+                    distances, indices = refine(
+                        dataset=item,
+                        queries=gpu_qfeatures,
+                        candidates=indices,
+                        k=cuml_alg_params["n_neighbors"],
+                        metric=cuml_alg_params["metric"],
+                    )
+
                 distances = cp.asarray(distances)
```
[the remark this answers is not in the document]
lijinf2: Yes, indices and distances returned from search API are already on the GPU. They are in Raft device_ndarray type.
cp.asarray is required to use cupy APIs to manipulate the device_ndarray. 
cp.asarray API doc says "If array is already on the device, no copy is performed".

python/src/spark_rapids_ml/knn.py
```diff
@@ -1504,35 +1526,57 @@ def _transform_internal(
...
-                    distances = distances * distances
+                # in case refine API reset inf distances to 0.
+                if cuml_alg_params["algorithm"] in {"ivf_pq", "ivfpq"}:
+                    distances[indices >= len(item)] = float("inf")
+
+                    # for the case top-1 nn got filled into indices
+                    top1_ind = indices[:, 0]
+                    rest_indices = indices[:, 1:]
+                    rest_distances = distances[:, 1:]
+                    rest_distances[rest_indices == top1_ind[:, cp.newaxis]] = float(
```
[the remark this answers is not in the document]
lijinf2: Yes, it is observed refine might return the top-1 index in other locations, if less than k items probed for a query. This can happen when the input candidate indices already contain many long_max (or top-1 index in other locations). 

Refine improves the indices and fills exact distances, but does not guarantee every query to probe at least k items.

python/tests/test_approximate_nearest_neighbors.py
lijinf2: Revised the comment in utils.py.

python/tests/utils.py
lijinf2: Revised to be more clear.

lijinf2:
build

python/src/spark_rapids_ml/knn.py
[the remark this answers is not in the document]
eordentlich: Shouldn't it return a permutation of the original top-k' indices for k' (  < k ) being the number of checked neighbors?   Why would there be duplicates?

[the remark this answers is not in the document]
lijinf2: Not sure if I follow you accurately. Yes, it is observed cuvs repeats the top-1 nn instead of permutating top k'.
At our test case test_return_fewer_k[float32-ivfpq-array], ivfpq + refine returns indices [2, 3, 2, 2] and distances [0., 0., 'inf', 'inf'] for data point of id 2 (and for data point of id 3 too since the two data points are identical).

lijinf2:
build

python/src/spark_rapids_ml/knn.py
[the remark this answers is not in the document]
lijinf2: Just added into the test case comparison with sg cuvs when less than k items probed. 

ivfflat: ensured output is the same as sg. 
ivfpq + refine: ensured the same behavior as ivfflat, and listening to the update of next release of sg.

lijinf2:
build

eordentlich (approved):
👍

PR #758, opened 2024-10-15
Spawn multiple pytest workers to speedup CI

lijinf2:
with pytest -n 4 on a V100 32GB GPU,
run_tests.sh got passed in 20 mins.
run_tests.sh --runslow got passed in 40 mins.
peak CPU memory usage was around 2GB.

### DIFF
--- a/python/run_test.sh
+++ b/python/run_test.sh
@@ -20,7 +20,7 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
 spark-submit --master local[1] tests_no_import_change/test_no_import_change.py 0.2
 
 echo "use --runslow to run all tests"
-pytest "$@" -n 3 benchmark/test_gen_data.py
-PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 3 --durations=10 tests
-#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 3 --durations=10 tests
+pytest "$@" -n 2 benchmark/test_gen_data.py
+PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 2 --durations=10 tests
+#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 2 --durations=10 tests
 #PYTHONPATH=`pwd`/benchmark pytest -ra "$@" --durations=10 tests_large
--- a/python/run_test.sh
+++ b/python/run_test.sh
@@ -20,7 +20,7 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
 spark-submit --master local[1] tests_no_import_change/test_no_import_change.py 0.2
 
 echo "use --runslow to run all tests"
-pytest "$@" -n 8 benchmark/test_gen_data.py
-PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 8 --durations=10 tests
-#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 8 --durations=10 tests
+pytest "$@" -n 3 benchmark/test_gen_data.py
+PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 3 --durations=10 tests
+#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 3 --durations=10 tests
 #PYTHONPATH=`pwd`/benchmark pytest -ra "$@" --durations=10 tests_large
--- a/python/run_test.sh
+++ b/python/run_test.sh
@@ -20,7 +20,7 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
 spark-submit --master local[1] tests_no_import_change/test_no_import_change.py 0.2
 
 echo "use --runslow to run all tests"
-pytest "$@" -n 4 benchmark/test_gen_data.py
-PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 4 --durations=10 tests
-#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 4 --durations=10 tests
+pytest "$@" -n 8 benchmark/test_gen_data.py
+PYTHONPATH=`pwd`/benchmark pytest -ra "$@" -n 8 --durations=10 tests
+#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow -n 8 --durations=10 tests
 #PYTHONPATH=`pwd`/benchmark pytest -ra "$@" --durations=10 tests_large
--- a/python/run_test.sh
+++ b/python/run_test.sh
@@ -19,8 +19,26 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
 # runs on cpu with spark-submit
 spark-submit --master local[1] tests_no_import_change/test_no_import_change.py 0.2
 
+
+# calculate pytest parallelism by following https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh
+MAX_PARALLEL=3
+NVIDIA_SMI_ARGS="" 
+if [ ${CUDA_VISIBLE_DEVICES} ]; then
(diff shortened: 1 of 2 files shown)


lijinf2:
build

lijinf2:
pre-merge CI took about 50 mins

eordentlich:
Wasn't it taking 50 min before?

eordentlich (commented):
@pxLi @YanxuanLiu Any potential issues with this?   
Seems to give 2x speedup in premerge.

lijinf2:
build

lijinf2:
build

pxLi:
> @pxLi @YanxuanLiu Any potential issues with this? Seems to give 2x speedup in premerge.

It may cause some GPU OOM or host memory OOM as in the CI pool, parts of machines have only 16/24GB GPU memory.

It would be nice for the ML, the test script could detect the gpu type and then dynamically set the parallelism like e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh#L113-L134 instead of hardcode the numbers thanks

lijinf2:
build

lijinf2:
build

python/run_test.sh
```diff
@@ -19,8 +19,17 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
 # runs on cpu with spark-submit
 spark-submit --master local[1] tests_no_import_change/test_no_import_change.py 0.2
 
+# calculate pytest parallelism following https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh
+GPU_MEM_PARALLEL=`nvidia-smi --query-gpu=memory.free --format=csv,noheader | awk '{if (MAX < $1){ MAX = $1}} END {print int((MAX - 2 * 1024) / ((1.5 * 1024) + 750))}'`
+CPU_CORES=`nproc`
+HOST_MEM_PARALLEL=`cat /proc/meminfo | grep MemAvailable | awk '{print int($2 / (8 * 1024 * 1024))}'`
```
eordentlich, line 25: Do these reflect container limits or VM limits?    And where did denominators come from?

lijinf2: denominator is inherited from the script that Peixin shared. The script expects 8GB peak memory usage during the whole testing. Ours was 2.5 GB I observed. 8GB is much larger than that so also works for us. 

@pxLi Do you have an idea on whether the the /proc/meminfo reflects container limits or VM limits?

lijinf2:
> > @pxLi @YanxuanLiu Any potential issues with this? Seems to give 2x speedup in premerge.
> 
> It may cause some GPU OOM or host memory OOM as in the CI pool, parts of machines have only 16/24GB GPU memory.
> 
> It would be nice for the ML, the test script could detect the gpu type (mem size) and then dynamically set the parallelism like e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh#L113-L134 instead of hardcode the numbers thanks

Thanks for sharing the link. 
Added the dynamic parallelism with a few modifications:
(1) calculate GPU_PARALLEL with the minimum device memory. 
(2) support CUDA_VISIBLE_DEVICES.

lijinf2:
build

lijinf2:
build

python/run_test.sh
lijinf2: Revised back to -n 3 instances.

lijinf2:
> > @pxLi @YanxuanLiu Any potential issues with this? Seems to give 2x speedup in premerge.
> 
> It may cause some GPU OOM or host memory OOM as in the CI pool, parts of machines have only 16/24GB GPU memory.
> 
> It would be nice for the ML, the test script could detect the gpu type (mem size) and then dynamically set the parallelism like e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh#L113-L134 instead of hardcode the numbers thanks

The script gave 9 on CI. It seems we need only 3 at this stage after discussing with Erik. No OOM was observed when I tested pre-merge and nightly tests locally. Should be safe.

python/run_test.sh
pxLi: > Do you have an idea on whether the the /proc/meminfo reflects container limits or VM limits?

the info is for local machine but not the correct value in docker. please give it a hard upper bound if possible.

like 

if dynamical parallelism > upper_bound then go with upper bound

pxLi:
> > > @pxLi @YanxuanLiu Any potential issues with this? Seems to give 2x speedup in premerge.
> > 
> > 
> > It may cause some GPU OOM or host memory OOM as in the CI pool, parts of machines have only 16/24GB GPU memory.
> > It would be nice for the ML, the test script could detect the gpu type (mem size) and then dynamically set the parallelism like e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh#L113-L134 instead of hardcode the numbers thanks
> 
> The script gave 9 on CI. It seems we need only 3 at this stage after discussing with Erik. No OOM was observed when I tested pre-merge and nightly tests locally. Should be safe.

I replied above, the more important is the GPU memory, feel free to remove the host memory detect part, just give GPU memory calculation (spark required ~1.5g GPU mem for each instance)
and put 3 as a upper bound for parallelism if possible thanks.

If you can also confirm hardcode 3 would not consume more than 15g (minimal GPU mem size in CI pool), we should be good too~

lijinf2:
build

lijinf2:
> > > > @pxLi @YanxuanLiu Any potential issues with this? Seems to give 2x speedup in premerge.
> > > 
> > > 
> > > It may cause some GPU OOM or host memory OOM as in the CI pool, parts of machines have only 16/24GB GPU memory.
> > > It would be nice for the ML, the test script could detect the gpu type (mem size) and then dynamically set the parallelism like e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-24.12/integration_tests/run_pyspark_from_build.sh#L113-L134 instead of hardcode the numbers thanks
> > 
> > 
> > The script gave 9 on CI. It seems we need only 3 at this stage after discussing with Erik. No OOM was observed when I tested pre-merge and nightly tests locally. Should be safe.
> 
> I replied above, the important part is the GPU memory, feel free to remove the host memory detect part, just give GPU memory calculation (e.g. spark test required ~1.5g GPU mem for each instance) and put 3 as a upper bound for parallelism if possible thanks.
> 
> If you can also confirm hardcode 3 would not consume more than 15g (minimal GPU mem size in CI pool), we should be good too~

@pxLi Got you. Just added back the dynamic parallelism, with host memory constraint removed yet a maximum parallelism 3 added. Our pre-merge and nightly tests took < 4 GB device memory tested in my workstation.  Hardcode 3 should consume no more than 12g.

python/run_test.sh
```diff
@@ -19,8 +19,26 @@ spark-rapids-submit --master local[1] tests_no_import_change/test_no_import_chan
...
+    else
+        TEST_PARALLEL=$TMP_PARALLEL
+fi
+echo "Test functions in benchmark/test_gen_data.py and tests/ directory will be executed in parallel with ${TEST_PARALLEL} pytest workers" 
+
 echo "use --runslow to run all tests"
-pytest "$@" benchmark/test_gen_data.py
-PYTHONPATH=`pwd`/benchmark pytest -ra "$@" --durations=10 tests
-#PYTHONPATH=`pwd`/benchmark pytest -ra --runslow --durations=10 tests
+pytest "$@" -n ${MAX_PARALLEL} benchmark/test_gen_data.py
```
eordentlich, line 41: Looks like TEST_PARALLEL not being used?    In that case delete the computation.

lijinf2:
build

python/run_test.sh
lijinf2: That's a typo. Revised to use TEST_PARALLEL. 

The computation can be helpful for device with limited memory. For instance, my workstation has a 2GB GPU and the computation helps reduce the parallelism to 1.

PR #759, opened 2024-10-15
Extend cuML parameter testing and cleanup

rishic3:
## Changes
`params.py`:
- Added logic to `_get_cuml_param` to handle the case where a cuML param is declared as a Spark param for algos w/out Spark equivalents.
  - Removes the need for identity mappings for UMAP, DBSCAN, and future cuML-only algos (except for cases where we override cuML's default value, as noted in `_param_mapping` docstring).
- Moved `DictTypeConverters` in from knn so that UMAP/future algos can leverage this function for Dict-type params.

`test_*.py`:
- Ensured all tests cover these checks:
  1. `Algorithm()._get_cuml_params_default() == get_default_cuml_parameters([CumlAlgorithm])`: Ensures that our internal dictionary of default cuML parameters matches those from cuML's API. 
  2. `assert_params(algorithm, spark_params, cuml_params)` with `default_params=True`: Ensures that we are correctly setting the default params in both the Spark parameter API and the internal cuML parameter dictionary.
  3. `assert_params(algorithm, spark_params, cuml_params)` with `default_params=False`: As above, checking that we correctly update params when non-default parameters are passed in.
  4. `assert algorithm.cuml_params == cuml_params`: An extra check to ensure the internal cuML dictionary fully matches the expected parameters under default and non-default conditions (catches any extra params without an equivalent in the other dict that may not have been caught in (iii)). 

**umap**:
- Added support for new params `build_algo` and `build_kwds` (and old param `metric_kwds`, which had been missing).
- Added a test for build_algo=nn_descent. 

**dbscan**:
- Unexposed `calc_core_sample_indices` from the user, since we [do not support it](https://github.com/NVIDIA/spark-rapids-ml/pull/598#discussion_r1548762983). 
- Updated documentation accordingly and added test to ensure it cannot be set. 

**algorithm docstrings**:
- Updated docstrings with default parameter values for consistency, fixed minor bugs/typos.

## Note about missing parameters 
In almost all algorithms, the corresponding cuML algorithm contains parameters that we do not support.  
Some unsupported params were "newly found" in this PR; this code currently excludes these parameters from the tests, (e.g., see [test_logistic_regression](https://github.com/rishic3/spark-rapids-ml/blob/param-tests/python/tests/test_logistic_regression.py#L190), [test_random_forest](https://github.com/rishic3/spark-rapids-ml/blob/param-tests/python/tests/test_random_forest.py#L128)), so that it can be merged.  

We may want do a pass over these excluded parameters to check if:
- the parameter is correctly omitted because we don't support it in the Spark version, or
- the parameter was added recently to cuML and we would like to update our algorithms to support it. 

If needed, in the latter case we can address parameters that should be supported in separate PR(s).

### DIFF
--- a/python/tests/test_kmeans.py
+++ b/python/tests/test_kmeans.py
@@ -63,14 +63,47 @@ def assert_centers_equal(
         assert a_center == pytest.approx(b_center, tolerance)
 
 
-def test_params() -> None:
+@pytest.mark.parametrize("default_params", [True, False])
+def test_params(default_params: bool) -> None:
     from cuml import KMeans as CumlKMeans
+
+    spark_params = {
+        "k": 2,
+        "initMode": "k-means||",
+        "tol": 0.0001,
+        "maxIter": 20,
+    }
 
     cuml_params = get_default_cuml_parameters(
         [CumlKMeans], ["handle", "output_type", "convert_dtype"]
     )
-    spark_params = KMeans()._get_cuml_params_default()
-    assert cuml_params == spark_params
+
+    # Ensure internal cuml defaults match actual cuml defaults
+    assert KMeans()._get_cuml_params_default() == cuml_params
+
+    cuml_params["n_clusters"] = 2  # cuml default gets overriden by spark default = 2
+    cuml_params["max_iter"] = 20  # cuml default gets overriden by spark default = 20
+    cuml_params["init"] = (
+        "k-means||"  # cuml default gets overriden by spark default = 'k-means||'
+    )
+
+    if default_params:
+        kmeans = KMeans()
+        seed = kmeans.getSeed()  # get the random seed that Spark generates
+        cuml_params["random_state"] = seed
+    else:
+        kmeans = KMeans(
+            n_clusters=10,
+            random_state=42,
+        )
+        cuml_params["n_clusters"] = 10
+        cuml_params["random_state"] = 42
+        spark_params["k"] = 10
+        spark_params["seed"] = 42
+
+    # Ensure both Spark API params and internal cuml_params are set correctly
+    assert_params(kmeans, spark_params, cuml_params)
+    assert kmeans.cuml_params == cuml_params
 
     # setter/getter
     from .test_common_estimator import _test_input_setter_getter
--- a/python/tests/test_kmeans.py
+++ b/python/tests/test_kmeans.py
@@ -66,12 +66,10 @@ def assert_centers_equal(
 @pytest.mark.parametrize("default_params", [True, False])
 def test_params(default_params: bool) -> None:
     from cuml import KMeans as CumlKMeans
+    from pyspark.ml.clustering import KMeans as SparkKMeans
 
     spark_params = {
-        "k": 2,
-        "initMode": "k-means||",
-        "tol": 0.0001,
-        "maxIter": 20,
+        param.name: value for param, value in SparkKMeans().extractParamMap().items()
     }
 
     cuml_params = get_default_cuml_parameters(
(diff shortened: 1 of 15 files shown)


rishic3:
build

python/tests/test_kmeans.py
```diff
@@ -63,14 +63,52 @@ def assert_centers_equal(
...
+    # Our algorithm overrides the following cuml parameters with their spark defaults:
+    spark_default_overrides = {
+        "n_clusters": 2,
+        "max_iter": 20,
+        "init": "k-means||",
+    }
+
+    for param, value in spark_default_overrides.items():
+        if param in cuml_params:
+            cuml_params[param] = value
```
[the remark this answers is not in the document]
rishic3: Not quite - if spark_default_overrides contains a key that doesn't exist in cuml_params, update() will add that key to cuml_params, which might miss failure cases, e.g. where cuML has deprecated a param.

rishic3:
build

rishic3:
build

rishic3:
build

rishic3:
build

rishic3:
build

rishic3:
build

python/tests/test_kmeans.py
```diff
@@ -63,14 +63,52 @@ def assert_centers_equal(
         assert a_center == pytest.approx(b_center, tolerance)
 
 
-def test_params() -> None:
+@pytest.mark.parametrize("default_params", [True, False])
+def test_params(default_params: bool) -> None:
     from cuml import KMeans as CumlKMeans
 
+    spark_params = {
+        "k": 2,
```
eordentlich, line 71: For algos in mllib can we get these from Spark somehow vs hard coding?

python/tests/test_kmeans.py
```diff
@@ -63,14 +63,52 @@ def assert_centers_equal(
...
-    assert cuml_params == spark_params
+
+    # Ensure internal cuml defaults match actual cuml defaults
+    assert KMeans()._get_cuml_params_default() == cuml_params
+
+    # Our algorithm overrides the following cuml parameters with their spark defaults:
+    spark_default_overrides = {
+        "n_clusters": 2,
+        "max_iter": 20,
+        "init": "k-means||",
```
eordentlich, line 88: Maybe better to set "init": spark_params["initMode"] and for the others.   Maybe even use the mapping between cuml names and spark names to set this.

python/tests/test_kmeans.py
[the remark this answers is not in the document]
eordentlich: If that happened wouldn't we catch it [here](https://github.com/NVIDIA/spark-rapids-ml/pull/759/files#diff-2e4ab3ba08dbaf41e0601b5a69fe8fffb83cf0e8764b679c59e7d02b6e6dac91R82)?

python/tests/test_kmeans.py
```diff
@@ -63,14 +63,52 @@ def assert_centers_equal(
...
+        if param in cuml_params:
+            cuml_params[param] = value
+
+    if default_params:
+        kmeans = KMeans()
+        seed = kmeans.getSeed()  # get the random seed that Spark generates
+        cuml_params["random_state"] = seed
+    else:
+        kmeans = KMeans(
+            n_clusters=10,
```
eordentlich, line 101: Seems for the overlapping algos with mllib we should use spark param names to initialize here (and below.)

python/tests/test_kmeans.py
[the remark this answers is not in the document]
rishic3: True - the scenario I mentioned won't happen; what can happen is that in special cases, a param might not appear in either cuML's or our default parameter dictionary (hence the assertion you mention will pass), but the param is then overridden and set in our algo. An example is `split_criterion` in [test_random_forest](https://github.com/rishic3/spark-rapids-ml/blob/param-tests/python/tests/test_random_forest.py#L160)—this param doesn't have a default in cuML, but it is mapped to "impurity", which does have a default in Spark MLLib. 

TLDR though, you're right, it's not an issue if we add a key that doesn't exist in cuml_params as long as our algo has the keys we expect in comparison.

python/tests/test_kmeans.py
rishic3: Good suggestion!

eordentlich:
build

eordentlich (approved):
Nice work!  👍

PR #767, opened 2024-10-23
update emr example and add initmode setter to kmeans

eordentlich:
rapids and emr 6.x are no longer compatible due to supported glibc version so updating to emr 7.3

There is also a work around to a [bug in spark 3.5.x](https://issues.apache.org/jira/browse/SPARK-49479) that got backported to emr 7.3 that prevents the emr 'step' wrapper from exiting after the spark job completes.   The work around is based on directly polling yarn on the spark application.   This requires enabling ssh access to the cluster.  

Also noticed as I was working on this that we are missing a setter for initMode in KMeans.   Unrelated to EMR but adding this easy fix to this PR.


eordentlich:
build

eordentlich:
build

lijinf2 (approved):
Looks good!

PR #769, opened 2024-10-28
Fix "No object to concat" Bug and support Some corner cases

lijinf2:
For both ANN and KNN:
(1) Support query_df is an empty dataframe

For ANN:
(1) Support item_df is an empty dataframe (root cause of triggering "no object to concat" error from pd.concat)

For KNN: 
(1) Select relevant columns (id, features) before calling union on query_df and item_df to avoid schema mismatch error


lijinf2:
build

eordentlich (approved):
👍

PR #771, opened 2024-10-30
Capture an assert failure from cuvs cagra and rephrase error message to include relevant params

lijinf2:
Error message: "k must be less than topk::kMaxCapacity (256)"
Captured with rephrased to f"cagra with ivf_pq build_algo expects intermediate_graph_degree ({intermediate_graph_degree}) to be smaller than 256"

Also added test cases of various type (float64, int16, int32).

### DIFF
--- a/python/src/spark_rapids_ml/knn.py
+++ b/python/src/spark_rapids_ml/knn.py
@@ -1099,6 +1099,12 @@ class ApproximateNearestNeighbors(
             "ivfpq",
             "cagra",
         }, "currently only ivfflat, ivfpq, and cagra are supported"
+        if not self._input_kwargs.get("float32_inputs", True):
+            get_logger(self.__class__).warning(
+                "This estimator supports only float32 inputs on GPU and will convert all other data types to float32. Setting float32_inputs to False will be ignored."
+            )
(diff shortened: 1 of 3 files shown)


lijinf2:
build

python/src/spark_rapids_ml/knn.py
```diff
@@ -1508,7 +1514,27 @@ def _transform_internal(
                 if isinstance(item, np.ndarray):
                     item = cp.array(item, dtype="float32")
 
-                index_obj = nn_object.build(build_params, item)
+                try:
+                    index_obj = nn_object.build(build_params, item)
+                except Exception as e:
+                    if "k must be less than topk::kMaxCapacity (256)" in str(e):
```
eordentlich, line 1520: Interesting way to do it.  What is advantage over  checking the error condition before invoking nn_object.build?

lijinf2: topk::kMaxCapacity (currently 256) is a global variable which may be updated to a higher or lower value some day. Hopefully we can catch the potential update by checking the error message after nn_object.build.

eordentlich (commented):
👍

eordentlich (approved):

PR #772, opened 2024-10-30
Enable UMAP to properly handle sparse data

rishic3:
Addressing [bug](https://nvbugspro.nvidia.com/bug/4937596) found by QA, where UMAP did not properly handle sparse inputs for 'jaccard' metric. 

#### Handling sparse data:
- Convert UMAP input data to CSR matrix in case of SparseVectors by inheriting enable_sparse_data_optim code.
- Return and store "raw_data" attribute in dict of lists containing the CSR attributes after fit. 
- Chunk and broadcast each CSR attribute independently, store result within dict.
- Fixed bug where UMAP was inheriting from the wrong CumlModel and was returning a new dataframe, rather than adding columns to the input dataframe. 
#### Reading/writing:
- Added logic to save/load CSR array components.
- Refactored to save model attributes as single compressed .npz file for all chunks (rather than separate .npy files) - requires compression time but less disk space. 
#### Testing:
- Added test for sparse inputs.
- Added test for model persistence in sparse case, including when data is saved/loaded in chunks. 

### Todos:
These features can be targeted for 24.12:
- Extend SparseDataGen to generate binary sparse dataset for testing jaccard metric (can be reused for exact/approx KNN)
- Refactor UMAP model persistence to enable saving/loading arrays to cloud storage (e.g., put arrays into a spark dataframe and use Spark dataframe writer)
- Chunking by bytes/row currently rests on the assumption of relatively uniform memory distribution - add better handling of skewed data.

### DIFF
--- a/python/benchmark/benchmark/bench_kmeans.py
+++ b/python/benchmark/benchmark/bench_kmeans.py
@@ -192,17 +192,6 @@ class BenchmarkKMeans(BenchmarkBase):
 
             cluster_centers = gpu_model.cluster_centers_
 
-            # temporary patch for DB with spark-rapids plugin
-            # this part is not timed so overhead is not critical, but should be reverted
-            # once https://github.com/NVIDIA/spark-rapids/issues/10770 is fixed
-            db_version = os.environ.get("DATABRICKS_RUNTIME_VERSION")
-            if db_version:
-                dim = len(cluster_centers[0])
-                # inject unsupported expr (slice) that is essentially a noop
-                df_for_scoring = df_for_scoring.select(
-                    F.slice(feature_col, 1, dim).alias(feature_col), output_col
-                )
-
         if num_cpus > 0:
             from pyspark.ml.clustering import KMeans as SparkKMeans
 
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -751,7 +751,8 @@ class _CumlCaller(_CumlParams, _CumlCommon):
                 concated_nnz = sum(triplet[0].nnz for triplet in inputs)  # type: ignore
                 if concated_nnz > np.iinfo(np.int32).max:
                     logger.warn(
-                        "the number of non-zero values of a partition is larger than the int32 index dtype of cupyx csr_matrix"
+                        f"The number of non-zero values of a partition exceeds the int32 index dtype of cupyx csr_matrix. \
+                        cupyx currently does not promote the dtype to int64 when concatenated; keeping as scipy to avoid overflow."
                     )
                 else:
                     inputs = [
--- a/python/src/spark_rapids_ml/core.py
+++ b/python/src/spark_rapids_ml/core.py
@@ -751,8 +751,9 @@ class _CumlCaller(_CumlParams, _CumlCommon):
                 concated_nnz = sum(triplet[0].nnz for triplet in inputs)  # type: ignore
                 if concated_nnz > np.iinfo(np.int32).max:
                     logger.warn(
-                        f"The number of non-zero values of a partition exceeds the int32 index dtype of cupyx csr_matrix. \
-                        cupyx currently does not promote the dtype to int64 when concatenated; keeping as scipy to avoid overflow."
+                        f"The number of non-zero values of a partition exceeds the int32 index dtype. \
+                        cupyx csr_matrix currently does not promote the dtype to int64 when concatenated; \
+                        keeping as scipy csr_matrix to avoid overflow."
                     )
                 else:
                     inputs = [
--- a/python/src/spark_rapids_ml/umap.py
+++ b/python/src/spark_rapids_ml/umap.py
@@ -1153,7 +1153,8 @@ class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
                 concated_nnz = sum(triplet[0].nnz for triplet in inputs)  # type: ignore
                 if concated_nnz > np.iinfo(np.int32).max:
                     logger.warn(
-                        "The number of non-zero values of a partition is larger than the int32 index dtype of cupyx csr_matrix."
+                        f"The number of non-zero values of a partition exceeds the int32 index dtype of cupyx csr_matrix. \
+                        cupyx currently does not promote the dtype to int64 when concatenated; keeping as scipy to avoid overflow."
                     )
                 else:
                     inputs = [
@@ -1302,78 +1303,48 @@ class UMAPModel(_CumlModelWithColumns, UMAPClass, _UMAPCumlParams):
 
         def _chunk_and_broadcast(
             sc: pyspark.SparkContext,
-            input: Union[np.ndarray, scipy.sparse.csr_matrix],
+            arr: np.ndarray,
             BROADCAST_LIMIT: int,
-        ) -> Union[
-            List[pyspark.broadcast.Broadcast],
-            Dict[str, List[pyspark.broadcast.Broadcast]],
-        ]:
+        ) -> List[pyspark.broadcast.Broadcast]:
             """
             Broadcast the input array, chunking it into smaller arrays if it exceeds the broadcast limit.
             """
-            if isinstance(input, np.ndarray):
-                total_bytes = input.nbytes
-                if total_bytes <= BROADCAST_LIMIT:
-                    return [sc.broadcast(input)]
+            if arr.nbytes < BROADCAST_LIMIT:
+                return [sc.broadcast(arr)]
 
-                rows_per_chunk = BROADCAST_LIMIT // (total_bytes // input.shape[0])
-                num_chunks = (input.shape[0] + rows_per_chunk - 1) // rows_per_chunk
-                return [
-                    sc.broadcast(input[i * rows_per_chunk : (i + 1) * rows_per_chunk])
-                    for i in range(num_chunks)
-                ]
-
-            elif isinstance(input, scipy.sparse.csr_matrix):
-                indices, indptr, data = input.indices, input.indptr, input.data
-                total_bytes = sum(arr.nbytes for arr in [indices, indptr, data])
-                if total_bytes <= BROADCAST_LIMIT:
-                    return {
-                        "indices": [sc.broadcast(indices)],
-                        "indptr": [sc.broadcast(indptr)],
-                        "data": [sc.broadcast(data)],
-                    }
-
-                rows_per_chunk = BROADCAST_LIMIT // (total_bytes // data.shape[0])
-                num_chunks = (data.shape[0] + rows_per_chunk - 1) // rows_per_chunk
-                # indptr is chunked differently because it has less rows; split to match num_chunks
-                indptr_rows_per_chunk = (
-                    indptr.shape[0] + (num_chunks - 1)
-                ) // num_chunks
-                return {
-                    "indices": [
-                        sc.broadcast(
-                            indices[i * rows_per_chunk : (i + 1) * rows_per_chunk]
-                        )
-                        for i in range(num_chunks)
-                    ],
-                    "indptr": [
-                        sc.broadcast(
-                            indptr[
-                                i
-                                * indptr_rows_per_chunk : (i + 1)
-                                * indptr_rows_per_chunk
-                            ]
-                        )
-                        for i in range(num_chunks)
-                    ],
-                    "data": [
-                        sc.broadcast(
-                            data[i * rows_per_chunk : (i + 1) * rows_per_chunk]
-                        )
-                        for i in range(num_chunks)
-                    ],
-                }  # NOTE: CSR chunks are not independently meaningful; do not use until recombined
-            else:
-                # should never get here
-                raise ValueError(f"Unsupported input type: {type(input)}")
+            rows_per_chunk = BROADCAST_LIMIT // (arr.nbytes // arr.shape[0])
+            if rows_per_chunk == 0:
+                raise ValueError(
+                    f"Array cannot be chunked into broadcastable pieces: \
+                        single row exceeds broadcast limit ({BROADCAST_LIMIT} bytes)"
+                )
+            num_chunks = (arr.shape[0] + rows_per_chunk - 1) // rows_per_chunk
+            return [
+                sc.broadcast(arr[i * rows_per_chunk : (i + 1) * rows_per_chunk])
+                for i in range(num_chunks)
+            ]
 
         spark = _get_spark_session()
         broadcast_embeddings = _chunk_and_broadcast(
             spark.sparkContext, self.embedding_, self.BROADCAST_LIMIT
         )
-        broadcast_raw_data = _chunk_and_broadcast(
-            spark.sparkContext, self.raw_data_, self.BROADCAST_LIMIT
-        )
+
+        if sparse_fit:
+            broadcast_raw_data = {
+                "indices": _chunk_and_broadcast(
+                    spark.sparkContext, self.raw_data_.indices, self.BROADCAST_LIMIT
+                ),
+                "indptr": _chunk_and_broadcast(
+                    spark.sparkContext, self.raw_data_.indptr, self.BROADCAST_LIMIT
+                ),
+                "data": _chunk_and_broadcast(
+                    spark.sparkContext, self.raw_data_.data, self.BROADCAST_LIMIT
+                ),
+            }
+        else:
+            broadcast_raw_data = _chunk_and_broadcast(
+                spark.sparkContext, self.raw_data_, self.BROADCAST_LIMIT
+            )
 
         def _construct_umap() -> CumlT:
             import cupy as cp
--- a/python/src/spark_rapids_ml/umap.py
+++ b/python/src/spark_rapids_ml/umap.py
@@ -958,31 +958,24 @@ class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
 
         pdf_output: PandasDataFrame = df_output.toPandas()
 
-        embeddings = np.array(
-            list(
-                pd.concat(
-                    [pd.Series(x) for x in pdf_output["embedding_"]], ignore_index=True
-                )
-            ),
-            dtype=np.float32,
-        )
-
         if self._sparse_fit:
+            embeddings = np.array(
+                list(
+                    pd.concat(
+                        [pd.Series(x) for x in pdf_output["embedding_"]],
+                        ignore_index=True,
+                    )
+                ),
+                dtype=np.float32,
+            )
             # Sparse data is returned in a dict of CSR format.
             indices = pdf_output["indices"].explode().to_numpy(dtype=np.int32)
             indptr = pdf_output["indptr"].explode().to_numpy(dtype=np.int32)
             data = pdf_output["data"].explode().to_numpy(dtype=np.float32)
             raw_data = {"indices": indices, "indptr": indptr, "data": data}
         else:
-            raw_data = np.array(
-                list(
-                    pd.concat(
-                        [pd.Series(x) for x in pdf_output["raw_data_"]],
-                        ignore_index=True,
-                    )
-                ),
-                dtype=np.float32,
-            )  # type: ignore
+            embeddings = np.vstack(pdf_output["embedding_"]).astype(np.float32)
+            raw_data = np.vstack(pdf_output["raw_data_"]).astype(np.float32)  # type: ignore
         del pdf_output
 
         def _chunk_and_broadcast(
@@ -1273,12 +1266,10 @@ class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
                     )
                 else:
                     yield pd.DataFrame(
-                        data=[
-                            {
-                                "embedding_": embedding[start:end].tolist(),
-                                "raw_data_": raw_data[start:end].tolist(),
-                            }
-                        ]
+                        {
+                            "embedding_": embedding[start:end].tolist(),
+                            "raw_data_": raw_data[start:end].tolist(),
+                        }
                     )
 
         output_df = dataset.mapInPandas(_train_udf, schema=self._out_schema())
@@ -1305,16 +1296,8 @@ class UMAP(UMAPClass, _CumlEstimatorSupervised, _UMAPCumlParams):
         else:
             return StructType(
                 [
-                    StructField(
-                        "embedding_",
-                        ArrayType(ArrayType(FloatType(), False), False),
-                        False,
-                    ),
-                    StructField(
-                        "raw_data_",
-                        ArrayType(ArrayType(FloatType(), False), False),
-                        False,
-                    ),
+                    StructField("embedding_", ArrayType(FloatType()), False),
+                    StructField("raw_data_", ArrayType(FloatType()), False),
                 ]
             )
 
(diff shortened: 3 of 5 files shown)


rishic3:
build

python/src/spark_rapids_ml/umap.py
```diff
@@ -894,6 +910,9 @@ def __init__(
         labelCol: Optional[str] = None,
         outputCol: Optional[str] = None,
         num_workers: Optional[int] = None,
+        enable_sparse_data_optim: Optional[
+            bool
+        ] = None,  # TBD: reuses 'vector -> sparse csr' conversion code, but will enable sparse data if first row is sparse for any metric. maybe set False by default, and None for 'jaccard'?
```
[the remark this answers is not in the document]
rishic3: It's only jaccard that [requires](https://github.com/rapidsai/cuml/blob/branch-24.10/python/cuml/cuml/manifold/umap_utils.pyx#L157) sparse inputs, but all other metrics can be used for both. I think None is the correct default here to allow sparse inputs for any metric.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1259,11 +1308,31 @@ def _construct_umap() -> CumlT:
...
-                else np.concatenate([chunk.value for chunk in driver_raw_data])
-            )
+
+            if self.sparse_fit:
+                if len(driver_raw_data) == 1:
+                    coo_list = driver_raw_data[0].value.tolist()
+                else:
+                    coo_list = np.concatenate(
+                        [chunk.value for chunk in driver_raw_data]
+                    ).tolist()
```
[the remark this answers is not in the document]
rishic3: Believe this avoids a concatenation (thus memory overhead of creating additional array buffer) for single chunk.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1141,19 +1179,26 @@ def _train_udf(pdf_iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
 
             num_sections = (len(embedding) + chunk_size - 1) // chunk_size
 
             for i in range(num_sections):
                 start = i * chunk_size
                 end = min((i + 1) * chunk_size, len(embedding))
+                if self._sparse_fit:
+                    # return sparse array as a list in COO format
+                    coo = raw_data[start:end].tocoo()
+                    raw_data = list(zip(coo.row, coo.col, coo.data))
```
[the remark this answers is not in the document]
rishic3: Good point. One option would be to return a list of dicts so that we can use a StructField and specify a separate type for each, but at the cost of memory overhead. 
Or we can return the row, col, and data lists as separate arrays and reconcatenate into COO LIL on the driver.

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
lijinf2: Got you. For jaccard, maybe we can add to docsting to explain only sparse is supported. Or it will be even better if we can check if inputs are sparse before calling cuml fit.

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
lijinf2: yeah, the idea sounds to work. Is it possible that we change the out_schema data type (whether float or int) depending on whether the cuml fit runs on dense or sparse matrix? out_schema is invoked after fit so we may be able to dynamically determine the output schema.

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
lijinf2: thought chunk.value is just a reference. I am not very sure though.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1283,15 +1354,29 @@ def _construct_umap() -> CumlT:
...
-            df: Union[pd.DataFrame, np.ndarray],
+            df: Union[pd.DataFrame, np.ndarray, scipy.sparse._csr.csr_matrix],
         ) -> pd.Series:
+            from cuml.common.sparse_utils import is_sparse
+
             embedding = umap.transform(df)
 
+            if is_sparse(df):
+                # scipy csr matrix to dense numpy array
+                df = df.toarray()  # type: ignore
```
rishic3, line 1371: TODO: there's definitely a better way to return the inputs rather than reconvert to dense array. Either SparseVectors or COO format would probably be preferable if possible.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1141,19 +1178,26 @@ def _train_udf(pdf_iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
             logger.info("Cuml fit complete")
 
             num_sections = (len(embedding) + chunk_size - 1) // chunk_size
 
             for i in range(num_sections):
                 start = i * chunk_size
                 end = min((i + 1) * chunk_size, len(embedding))
+                if use_sparse_array:
+                    # return sparse array as a list in COO format
+                    coo = raw_data[start:end].tocoo()
```
eordentlich, line 1191: Any reason to not stick with csr format here?  consistent representation like output of unwrap_udt on sparse vector.   And change the outschema (add an extra column for the indices) as @lijinf2 suggests.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1118,18 +1142,31 @@ def _train_udf(pdf_iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
                 else:
                     features = np.array(list(pdf[alias.data]), order=array_order)
-                # experiments indicate it is faster to convert to numpy array and then to cupy array than directly
-                # invoking cupy array on the list
                 if cuda_managed_mem_enabled:
-                    features = cp.array(features)
+                    if use_sparse_array:
+                        concated_nnz = sum(triplet[0].nnz for triplet in inputs)  # type: ignore
+                        if concated_nnz > np.iinfo(np.int32).max:
+                            logger.warn(
```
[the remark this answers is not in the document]
rishic3: If the 32 bit limit is exceeded the features will be kept as dense arrays, which I think should be the right behavior rather than throwing an error. In the case of 'jaccard' we'll throw an error once the dense features reach the fit func.

python/src/spark_rapids_ml/umap.py
rishic3: I think the advantage of COO is that since the row_indices, col_indices, and values all have the same shape, we can stack these into a numpy array to avoid storing a dense representation internally in the model. With CSR we can't stack into a numpy array, and we'd have to either convert to dense or to COO and do the same stacking trick.

eordentlich: Isn't COO redundant wrt to CSR in that it repeats the row index?   Could we use scipy's sparse matrix writing/reading methods to do the writes in the sparse case?   seems like these manipulations are reinventing the wheel in some sense and also wind up making multiple copies of things.   How to broadcast a scipy csr or the individual index and value columns though seems less clear.   Doubt Spark knows how serialize it directly but possible it does.   By the way Spark has its own SparseMatrix https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/linalg.html#SparseMatrix which looks to be CSC (not CSR) internally, though there is a transpose bit.  No write method though so probably not good in this case.

python/tests/test_umap.py
```diff
@@ -493,3 +493,71 @@ def assert_umap_model(model: UMAPModel) -> None:
...
+
+    with CleanSparkSession() as spark:
+        data = []
+        for i in range(n_rows):
+            # Generate binary sparse data compatible with Jaccard, with nnz non-zero values per row.
+            indices = [(i + j) % n_cols for j in range(nnz)]
+            values = [1] * nnz
+            sparse_vector = SparseVector(n_cols, dict(zip(indices, values)))
+            data.append((sparse_vector,))
+        df = spark.createDataFrame(data, ["features"])
```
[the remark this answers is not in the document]
lijinf2: Is it possible to extend SparseDataGen to generate dataset for testing jaccard distances? That will be helpful to other algorithms such as exact/appro nearest neighbors search.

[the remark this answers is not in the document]
rishic3: Added to TODOS to target in 24.12.

rishic3:
build

rishic3:
build

rishic3:
build

python/benchmark/benchmark/bench_kmeans.py
```diff
@@ -192,17 +192,6 @@ def gpu_cache_df(df: DataFrame) -> DataFrame:
...
-            # this part is not timed so overhead is not critical, but should be reverted
-            # once https://github.com/NVIDIA/spark-rapids/issues/10770 is fixed
-            db_version = os.environ.get("DATABRICKS_RUNTIME_VERSION")
-            if db_version:
-                dim = len(cluster_centers[0])
-                # inject unsupported expr (slice) that is essentially a noop
-                df_for_scoring = df_for_scoring.select(
-                    F.slice(feature_col, 1, dim).alias(feature_col), output_col
-                )
-
```
rishic3, line 205: Removing since it looks resolved: https://github.com/NVIDIA/spark-rapids/issues/10770

python/src/spark_rapids_ml/umap.py
```diff
@@ -1259,11 +1423,32 @@ def _construct_umap() -> CumlT:
...
+
+            if sparse_fit:
+                if not isinstance(driver_raw_data, dict):
+                    raise ValueError("Expected raw data as a CSR dict for sparse fit.")
+                indices = np.concatenate(
+                    [chunk.value for chunk in driver_raw_data["indices"]]
+                )
+                indptr = np.concatenate(
+                    [chunk.value for chunk in driver_raw_data["indptr"]]
+                )
```
lijinf2, line 1435: Can we simply concatenate indptr here? 
Would it be possible that indptr_a = [0, 4], indptr_b=[0, 2, 3] and the indptr_concate should be [0, 4, 6, 7]?

python/tests/test_umap.py
```diff
@@ -493,3 +549,50 @@ def assert_umap_model(model: UMAPModel) -> None:
...
         assert trust_diff <= 0.15
+
+
+@pytest.mark.parametrize("n_rows", [3000])
+@pytest.mark.parametrize("n_cols", [64, 128])
+@pytest.mark.parametrize("nnz", [7, 12])
+def test_umap_sparse_vector(
+    n_rows: int, n_cols: int, nnz: int, gpu_number: int, tmp_path: str
+) -> None:
+    import pyspark
```
lijinf2, line 560: Better to add a new test case where a csr_matrix gets split in estimator class and reconstructed in model class.

python/src/spark_rapids_ml/umap.py
```diff
@@ -946,45 +966,110 @@ def _fit(self, dataset: DataFrame) -> "UMAPModel":
...
+                        )
+                        for i in range(num_chunks)
+                    ],
+                    "data": [
+                        sc.broadcast(
+                            input["data"][i * rows_per_chunk : (i + 1) * rows_per_chunk]
+                        )
+                        for i in range(num_chunks)
+                    ],
+                }  # NOTE: CSR chunks are not independently meaningful; do not use until recombined
```
lijinf2, line 1052: Nit: we should be able to combine the duplicate code with a for loop like "for key in ["indices", "indptr", "data"]".

python/src/spark_rapids_ml/umap.py
rishic3: Do you mean concatenating on the dictionary value directly? The list comprehension is needed to to retrieve the array from the broadcast object.
Each of the CSR components are one-dim column vectors and are chunked by row; np.concatenate will by default always concatenate by row and the list will preserve the order.

python/tests/test_umap.py
rishic3: This was added to the model persistence test.

python/src/spark_rapids_ml/umap.py
rishic3: Wrote it this way since indptr requires a different chunking index. indices and data code is indeed repetitive but this might just be clearer.

python/tests/test_umap.py
```diff
@@ -493,3 +549,50 @@ def assert_umap_model(model: UMAPModel) -> None:
         trust_diff = loc_umap - dist_umap
 
         assert trust_diff <= 0.15
+
+
+@pytest.mark.parametrize("n_rows", [3000])
+@pytest.mark.parametrize("n_cols", [64, 128])
+@pytest.mark.parametrize("nnz", [7, 12])
+def test_umap_sparse_vector(
+    n_rows: int, n_cols: int, nnz: int, gpu_number: int, tmp_path: str
```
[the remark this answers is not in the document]
rishic3: Added the most commonly used metrics for binary sparse data.

python/src/spark_rapids_ml/umap.py
lijinf2: Got you. Yeah, need to ensure the values of indptr arrays are not modified by chunking.

rishic3:
Re: discussion on how to avoid chunking twice - chunking when returning from fit is by # rows, and is meant to help conform with maxRecordsPerBatch so that Spark only loads a batch at a time into memory, whereas chunking on the driver side is based on # bytes to stay below the 8GB broadcast limit.

During fit we can calculate the tighter of the two restrictions prior to yielding to the driver and then reuse these chunks during broadcast. Though if the maxRecordsPerBatch chunk is significantly smaller in bytes, we'd have the overhead of unnecessarily broadcasting lots of small chunks. Not sure if there's a fix that makes sense here.

rishic3:
> Re: discussion on how to avoid chunking twice - chunking when returning from fit is by # rows, and is meant to help conform with maxRecordsPerBatch so that Spark only loads a batch at a time into memory, whereas chunking on the driver side is based on # bytes to stay below the 8GB broadcast limit.
> 
> During fit we can calculate the tighter of the two restrictions prior to yielding to the driver and then reuse these chunks during broadcast. Though if the maxRecordsPerBatch chunk is significantly smaller in bytes, we'd have the overhead of unnecessarily broadcasting lots of small chunks. Not sure if there's a fix that makes sense here.

Also, chunking the CSR matrix by # rows when returning from _fit poses separate challenges since the CSR components and the embedding matrix will have different # rows. It's unclear the number of chunks we should return here. 
If we could chunk the CSR matrix directly and reconcatenate using scipy.sparse.vstack that would be ideal, but there's no easy way to get the CSR matrix into a serializable format (e.g., python lists) unless we extract the components. Will think about this more.

python/src/spark_rapids_ml/umap.py
```diff
@@ -946,45 +966,110 @@ def _fit(self, dataset: DataFrame) -> "UMAPModel":
...
+                raise ValueError(f"Unsupported input type: {type(input)}")
 
         spark = _get_spark_session()
-        broadcast_embeddings = [
-            spark.sparkContext.broadcast(chunk) for chunk in _chunk_arr(embeddings)
-        ]
-        broadcast_raw_data = [
-            spark.sparkContext.broadcast(chunk) for chunk in _chunk_arr(raw_data)
-        ]
+        broadcast_embeddings = _chunk_and_broadcast(
```
eordentlich, line 1059: I know this was from before, but it doesn't feel right to have the attributes of the model be broadcast objects vs just the underlying objects.  The broadcast wrapper is then applied in those cases where actual broadcast takes place, like just before transform.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1396,14 +1593,46 @@ def load(self, path: str) -> "_CumlEstimator":
...
+                            key=lambda x: int(x.split("_")[1]),
+                        )
+                        return sorted_keys
+
+                    indices_keys = _get_sorted_data_keys(npz_data, "indices")
+                    indptr_keys = _get_sorted_data_keys(npz_data, "indptr")
+                    data_keys = _get_sorted_data_keys(npz_data, "data")
+                    raw_data_dict = {
+                        "indices": [
+                            spark.sparkContext.broadcast(npz_data[key])
```
eordentlich, line 1618: Related to above comment, seems odd to create broadcast upon loading the model from store.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1201,36 +1338,62 @@ def _pre_process_data(
...
+        """
+        Returns the raw data used to fit the model. If the input data was sparse, this will be a scipy csr matrix.
+        """
+        if isinstance(self.raw_data_, dict):
+            indices = []
+            indptr = []
+            data = []
+            for i, chunk in enumerate(self.raw_data_["indices"]):
+                indices.extend(chunk.value)
+                indptr.extend(self.raw_data_["indptr"][i].value)
```
eordentlich, line 1385: there would be no need to take value here for example if underlying chunks weren't yet broadcast.

python/src/spark_rapids_ml/umap.py
```diff
@@ -946,45 +966,110 @@ def _fit(self, dataset: DataFrame) -> "UMAPModel":
...
-                pd.concat(
-                    [pd.Series(x) for x in pdf_output["raw_data_"]], ignore_index=True
-                )
-            ),
-            dtype=np.float32,
-        )
+
+        if self._sparse_fit:
+            # Sparse data is returned in a dict of CSR format.
+            indices = pdf_output["indices"].explode().to_numpy(dtype=np.int32)
```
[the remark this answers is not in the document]
eordentlich: Explode seems costly, but not sure.

python/src/spark_rapids_ml/umap.py
eordentlich: Whether to store attributes as chunked or not is not completely clear, but storing as broadcast seems off.

python/benchmark/benchmark/bench_kmeans.py
eordentlich: Good find.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1137,26 +1232,54 @@ def _train_udf(pdf_iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
...
+                                "embedding_": embedding[start:end].tolist(),
+                                "indices": indices.tolist(),
+                                "indptr": indptr.tolist(),
+                                "data": data.tolist(),
+                            }
+                        ]
+                    )
+                else:
+                    yield pd.DataFrame(
+                        data=[
```
rishic3, line 1276: The data = [...] is a bug as it is packing everything into a single row so the whole thing is one Arrow record. Also explains why the CSR components with mismatched sizes were passing tests despite mismatched sizes.

rishic3:
build

python/src/spark_rapids_ml/umap.py
```diff
@@ -1201,36 +1247,47 @@ def _pre_process_data(
...
             embedding_=embedding_,
             raw_data_=raw_data_,
+            sparse_fit=sparse_fit,
             n_cols=n_cols,
             dtype=dtype,
         )
         self.embedding_ = embedding_
         self.raw_data_ = raw_data_
+        self._sparse_fit = sparse_fit  # If true, raw data is a sparse CSR matrix
+        self.BROADCAST_LIMIT = 8 << 30  # Spark broadcast limit: 8GiB
```
lijinf2, line 1272: Would this limit be affected by spark configuration parameters such as spark.driver.maxResultSize? Maybe create a ticket if we this needs to be investigated.

lijinf2:
Thanks for addressing the comments. I am good with merging the PR. Some corner cases can be studied in the future.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1240,9 +1297,83 @@ def _get_cuml_transform_func(
...
+                indptr_rows_per_chunk = (
+                    indptr.shape[0] + (num_chunks - 1)
+                ) // num_chunks
+                return {
+                    "indices": [
+                        sc.broadcast(
+                            indices[i * rows_per_chunk : (i + 1) * rows_per_chunk]
+                        )
+                        for i in range(num_chunks)
+                    ],
```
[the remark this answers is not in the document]
rishic3: For the second case:
rows_per_chunk = BROADCAST_LIMIT // (total_bytes // len(data)) = BROADCAST_LIMIT // (36 // 4)
=> num_chunks = (4 + 1 - 1) // 1 = 4, so it would work.

For the first case, agreed we definitely need to check the case where data is sufficiently dense that BROADCAST_LIMIT < bytes-per-row. 

In general agree that data skewness is an issue and it's in the things-to-address for 24.12. The current code overestimates the size of each row by using total_bytes // len(data) rather than data.nbytes // len(data) to give some buffer room, but this is kind of arbitrary and needs a revision. 
A simple approach is to check the nbytes of any outgoing chunks to ensure nothing exceeds the limit, and rechunk in a smart way to get under the boundary. Or as Erik mentioned, just reusing the Arrow record batches that were returned from fit, and applying a double-check on the sizes of these. 

test_umap_chunking explicitly tests num_chunks > 1. But certainly it could apply more corner cases, e.g. very large row density and skewness.

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
rishic3: If you're suggesting that it changes dynamically depending on other confs the answer should be no, as it's [hardcoded in](https://github.com/apache/spark/blob/e17d8ecabcad6e84428752b977120ff355a4007a/sql/core/src/main/scala/org/apache/spark/sql/execution/exchange/BroadcastExchangeExec.scala#L226); if driver.maxResultSize < broadcast_limit that would be an issue, but we warn the user to use sufficient driver memory in [the notebooks](https://github.com/NVIDIA/spark-rapids-ml/blob/5eb2ca8673c8804a06cbc68fb3b6383d2d3e4d31/notebooks/umap.ipynb).

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
eordentlich: Where does the data convert back to dense?

rishic3:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -751,7 +751,8 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
                 concated_nnz = sum(triplet[0].nnz for triplet in inputs)  # type: ignore
                 if concated_nnz > np.iinfo(np.int32).max:
                     logger.warn(
-                        "the number of non-zero values of a partition is larger than the int32 index dtype of cupyx csr_matrix"
+                        f"The number of non-zero values of a partition exceeds the int32 index dtype of cupyx csr_matrix. \
+                        cupyx currently does not promote the dtype to int64 when concatenated; keeping as scipy to avoid overflow."
```
lijinf2, line 755: "cupyx" -> "cupyx csr_matrix", "scipy" - > "scipy csr_matrix" to be more accurate?

python/src/spark_rapids_ml/umap.py
```diff
@@ -1240,9 +1298,53 @@ def _get_cuml_transform_func(
         Optional[_EvaluateFunc],
     ]:
         cuml_alg_params = self.cuml_params
-        driver_embedding = self.embedding_
-        driver_raw_data = self.raw_data_
-        outputCol = self.getOutputCol()
+        sparse_fit = self._sparse_fit
+        n_cols = self.n_cols
+
+        def _chunk_and_broadcast(
```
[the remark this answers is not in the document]
lijinf2: Looks good!

lijinf2 (approved):

lijinf2:
build

PR #773, opened 2024-11-05
address python workers not exiting properly after failures

eordentlich:
This eliminates zombie workers on my workstation when I try to run the benchmark script on my two gpu cards, only one of which supports the app.   The bad gpu worker fails and exits, while, before this change, the good gpu python worker would hang forever.   With this change, the good worker also exists.

Would be good to test other scenarios.


eordentlich:
build

lijinf2 (approved):

wbo4958:
Looks like this case could also be in XGBoost case, right? @eordentlich

PR #785, opened 2024-11-12
doc patches and notebook for no import change

eordentlich:
demos how to do no import change in a notebook with one global import in the first cell.
In future prs will give examples, where possible in various csps, how to do that one import when kernels are launched instead of in the notebook.


eordentlich:
build

lijinf2 (approved):
Looks good!

PR #787, opened 2024-11-20
fix rmm managed memory resource initialization to resolve some intermittent …

eordentlich:
…memory issues

Looks like rmm.reinitialize destroys memory resources assigned to all other devices besides the current device.  This way of initializing as in this PR and doing only once per process avoids this.  I think the intermittent cuda errors were due to hit or miss uses of destroyed resources on the c++ api rmm operations.   So this should address the issue.

Also, in umap, current device was set before the memory resource was initialized.  Needs to be the other way around.

also pin numpy < 1 in readme

### DIFF
--- a/python/src/spark_rapids_ml/utils.py
+++ b/python/src/spark_rapids_ml/utils.py
@@ -144,6 +144,21 @@ def _get_gpu_id(task_context: TaskContext) -> int:
     return gpu_id
 
 
+def _configure_memory_resource(uvm_enabled: bool = False) -> None:
+    import cupy as cp
+    import rmm
+    from rmm.allocators.cupy import rmm_cupy_allocator
+
+    if uvm_enabled:
+        if not type(rmm.mr.get_current_device_resource()) == type(
+            rmm.mr.ManagedMemoryResource()
+        ):
+            rmm.mr.set_current_device_resource(rmm.mr.ManagedMemoryResource())
+
+        if not cp.cuda.get_allocator().__name__ == rmm_cupy_allocator.__name__:
+            cp.cuda.set_allocator(rmm_cupy_allocator)
+
+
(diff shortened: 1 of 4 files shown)


eordentlich:
build

eordentlich:
build

python/src/spark_rapids_ml/utils.py
```diff
@@ -144,6 +144,21 @@ def _get_gpu_id(task_context: TaskContext) -> int:
...
+def _configure_memory_resource(uvm_enabled: bool = False) -> None:
+    import cupy as cp
+    import rmm
+    from rmm.allocators.cupy import rmm_cupy_allocator
+
+    if uvm_enabled:
+        if not type(rmm.mr.get_current_device_resource()) == type(
+            rmm.mr.ManagedMemoryResource()
+        ):
+            rmm.mr.set_current_device_resource(rmm.mr.ManagedMemoryResource())
```
lijinf2, line 156: Looks good.

One question: any idea on the difference of this line compared with "rmm.reinitialize(managed_memory=True)"?

lijinf2 (approved):

python/src/spark_rapids_ml/utils.py
eordentlich: Yes.   rmm.reinitialize also resets memory resources for previously specified devices to default mr's.   This I believe was the source of our problems.  It destroyed the original memory resources, replacing the device to mr mapping to point to the new defaults.  However this is only done on the python side.  On the c++ side, the corresponding map still points to the destroyed mrs (which are the same object), leading to dangling pointers.   So this would lead to undefined behavior in the C++ rmm invocations, if the devices were changing in the same process, like between fit and transform, with python worker reuse on.   The observations and code inspection were consistent with the above.

PR #790, opened 2024-11-27
update etl plugin to 24.10 to pull in some benchmark related fixes

eordentlich:
and turn off mlflow autlogging for benchmarks which can be non-trivially resource intensive.

This will pull the fix to the etl plugin that allows kmeans benchmark metrics to run without hang.


eordentlich:
build

lijinf2 (approved):
Looks good!

PR #799, opened 2024-12-09
Add a test case for sparse logistic regression with standardization per a hanging issue has been fixed in cuml 24.12

lijinf2:
The hanging occurs when only some of all GPUs get sparse matrix of all zeroes.

### DIFF
--- a/python/tests/test_logistic_regression.py
+++ b/python/tests/test_logistic_regression.py
@@ -2266,11 +2266,8 @@ def test_sparse_all_zeroes(
         compare_model(gpu_model, cpu_model, bdf)
 
 
-@pytest.mark.slow
-# @pytest.mark.parametrize("standardization", [True, False])
-# @pytest.mark.parametrize("fit_intercept", [True, False])
 @pytest.mark.parametrize("standardization", [True])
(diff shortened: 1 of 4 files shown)


lijinf2:
build

python/tests/test_logistic_regression.py
```diff
@@ -2264,3 +2264,48 @@ def test_sparse_all_zeroes(
...
+) -> None:
+    tolerance = 0.001
+
+    if gpu_number < 2:
+        pytest.skip(reason="test_sparse_one_gpu_zeroes requires at least 2 GPUs")
+    gpu_number = 2
+
+    with CleanSparkSession() as spark:
+        data = [
+            Row(label=1.0, weight=1.0, features=Vectors.sparse(2, {0: 10.0, 1: 20.0})),
```
eordentlich, line 2284: weight column needed?

eordentlich (approved):
👍

python/tests/test_logistic_regression.py
lijinf2: Good catch. Will remove it in the next PR.

PR #807, opened 2024-12-12
Fix a bug in copy() of LogisticRegression that does not infer the penalty cuml parameter

lijinf2:


lijinf2:
> Can you check if there is a more general solution that invokes code/logic that does this for initialization? In that case, it may resolve similar issues in the future and maybe with other classes?

Revised the PR to make the function independent of LogisticRegression and its parameters. Let me know what you think.

lijinf2:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -1163,6 +1163,16 @@ def _set_params(self, **kwargs: Any) -> "LogisticRegression":
             self._set_cuml_reg_params()
         return self
 
+    def copy(
```
eordentlich, line 1166: Is there a way to move this to base class in core and somehow leverage _set_params which already calls set_cuml_reg_params().  Or refactor _set_params a bit to enable this?

lijinf2: Good suggestion. Revised to have this moved to base class in params.py.

lijinf2:
build

lijinf2:
@eordentlich The PR has been updated to address comments and includes test cases for all estimators. Please review and let me know if you have any further feedback.

Currently, the `copy()` function works for all initialization parameters except `float32_inputs` and `num_workers`. These need to be converted to Spark Params, and I will make that change if the overall PR looks acceptable

python/src/spark_rapids_ml/tree.py
```diff
@@ -148,8 +154,57 @@ class _RandomForestCumlParams(
     HasFeaturesCols,
     HasLabelCol,
 ):
+
+    n_streams = Param(
```
eordentlich, line 158: Are these converted to Params to enable modification on copy?   For spark.mllib estimators I think ok to do that only for mapped params.   Or at least is a separate topic.

python/tests/test_logistic_regression.py
```diff
@@ -329,6 +329,88 @@ def test_params(tmp_path: str, caplog: LogCaptureFixture) -> None:
...
+            None,
+        ),
+        (
+            LogisticRegression,
+            {"verbose": True},
+            {"verbose": True},
+        ),
+    ],
+)
+def test_copy(
```
eordentlich, line 382: this is used in other tests so seems better to move in a generic central place like utils?

python/src/spark_rapids_ml/params.py
```diff
@@ -128,6 +128,36 @@ def _ensureIdCol(self, df: DataFrame) -> DataFrame:
...
+        raise TypeError("Could not convert %s to Union[int, bool]" % value)
+
+
+class HasVerboseParam(Params):
+    """
+    Parameter to enable displaying verbose messages from cuml.
+    Refer to the cuML documentation for details on verbosity levels.
+    """
+
+    verbose: "Param[Union[int, bool]]" = Param(
```
eordentlich, line 149: Why does verbose need to be a spark param?

python/src/spark_rapids_ml/tree.py
lijinf2: Yeah, this is to avoid changing function signature: def copy(self: P, extra: Optional["ParamMap"] = None) -> P, and ParamMap = Dict[pyspark.ml.param.Param, Any]. 

copy() only accepts Param type.

python/src/spark_rapids_ml/params.py
lijinf2: Same as above. copy only accepts Param type in the argument. Not sure if it is a good idea to change the function signature or mute the mypy type checking on the function.

python/src/spark_rapids_ml/tree.py
rishic3: Sorry if I'm missing something -  how come we need to declare RF cuml params as Params here and not in other algos?

python/src/spark_rapids_ml/params.py
```diff
@@ -273,21 +303,18 @@ def copy(self: P, extra: Optional["ParamMap"] = None) -> P:
         instance: P = super().copy(extra)
         cuml_params = instance.cuml_params.copy()
 
+        instance._cuml_params = cuml_params
```
rishic3, line 306: For future reference maybe add a docstring for copy()?

python/src/spark_rapids_ml/params.py
eordentlich: Ok.  For this and above let's make sure it the changes don't break anything.   copy() with extra params is only used in crossvalidator and changing verbose wouldn't be a part of that and we can ignore non-spark params for algo's in spark mllib.   
But if we're sure won't break anything, seems ok I guess to include these.

Otherwise, looks great.

rishic3:
Not directly related to this PR but should DBSCAN setters/getters for cuml Params be placed in _DBSCANCumlParams class rather than DBSCAN to align with UMAP/KNN?

python/src/spark_rapids_ml/tree.py
lijinf2: Right. We will need to add the cuml-only params (e.g. `whiten` and `svd_solver`) one by one as Param, in order to get them working properly.

python/src/spark_rapids_ml/params.py
lijinf2: added

lijinf2:
> Not directly related to this PR but should DBSCAN setters/getters for cuml Params be placed in _DBSCANCumlParams class rather than DBSCAN to align with UMAP/KNN?

"Thank you for sharing your thoughts! I have updated the PR accordingly. To ensure proper functionality in copy(), any cuML-only parameter must be declared as a pyspark.Param type. Please take another look when you have a moment.

Regarding the alignment of DBSCAN setters/getters with UMAP/KNN, it seems we may need to review other estimators as well. Perhaps we can create a ticket to track this. Similarly, we should track the broader issue of supporting all cuML-only parameters (e.g., 'whiten' in PCA)."

rishic3:
LGTM. Thanks @lijinf2!

python/tests/test_kmeans.py
```diff
@@ -183,6 +183,22 @@ def test_kmeans_params(
     assert not kmeans_float32._float32_inputs
 
 
+def test_copy() -> None:
```
rishic3, line 186:
```suggestion
def test_kmeans_copy() -> None:
```

lijinf2:
build

eordentlich (approved):
👍

PR #823, opened 2025-01-13
Avoid local dir creation, fix dense ordering

rishic3:


rishic3:
build

python/src/spark_rapids_ml/umap.py
```diff
@@ -1473,11 +1474,19 @@ def write_sparse_array(array: scipy.sparse.spmatrix, df_dir: str) -> None:
...
-            data_df = spark.createDataFrame(pd.DataFrame(array), schema=schema)
+            data_df = spark.createDataFrame(
+                pd.DataFrame(
+                    {
+                        "row_id": range(array.shape[0]),
+                        "data": array.tolist(),
+                    }
+                ),
+                schema=schema,
+            )
```
eordentlich, line 1489: Below and elsewhere in this class is it correct to use overwrite when writing?   This might be counter to the overwrite MLWriter api.    If that is not invoked, a user would not expect overwrite to be allowed.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1547,8 +1554,8 @@ def read_sparse_array(
             return scipy.sparse.csr_matrix((data, indices, indptr), shape=csr_shape)
 
         def read_dense_array(df_path: str) -> np.ndarray:
-            data_df = spark.read.parquet(df_path)
-            return np.array(data_df.collect(), dtype=np.float32)
+            data_df = spark.read.parquet(df_path).orderBy("row_id")
+            return np.array([row.data for row in data_df.collect()], dtype=np.float32)
```
eordentlich, line 1558: toPandas might be better here followed by np.array(list(data_pandas_df.data))

python/src/spark_rapids_ml/umap.py
```diff
@@ -1495,8 +1504,6 @@ def write_dense_array(array: np.ndarray, df_path: str) -> None:
         assert model_attributes is not None
 
         data_path = os.path.join(path, "data")
-        if not os.path.exists(data_path):
```
eordentlich, line 1498: Would be good to have a test that checks for expected files and directories?

python/src/spark_rapids_ml/umap.py
```diff
@@ -1547,8 +1554,8 @@ def read_sparse_array(
             return scipy.sparse.csr_matrix((data, indices, indptr), shape=csr_shape)
 
         def read_dense_array(df_path: str) -> np.ndarray:
-            data_df = spark.read.parquet(df_path)
-            return np.array(data_df.collect(), dtype=np.float32)
+            data_df = spark.read.parquet(df_path).orderBy("row_id")
```
eordentlich, line 1557: I wonder if there is test for the order, one that would fail if orderby was omitted.

python/src/spark_rapids_ml/umap.py
```diff
@@ -1473,11 +1474,19 @@ def write_sparse_array(array: scipy.sparse.spmatrix, df_dir: str) -> None:
...
+                    StructField("row_id", LongType(), False),
+                    StructField("data", ArrayType(FloatType(), False), False),
                 ]
             )
-            data_df = spark.createDataFrame(pd.DataFrame(array), schema=schema)
+            data_df = spark.createDataFrame(
+                pd.DataFrame(
+                    {
+                        "row_id": range(array.shape[0]),
+                        "data": array.tolist(),
```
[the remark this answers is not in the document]
rishic3: While list(array) is more efficient since tolist() does a deep conversion of each row to python lists, Spark will throw an error with list(array) if `spark.sql.execution.arrow.pyspark.enabled=false`, since pyarrow would no longer handle the numpy -> arrow array conversion.

python/src/spark_rapids_ml/umap.py
rishic3: A multi-gpu env (e.g., DGX) where Spark's default parallelism is >1 would have caught it and I should have tested there with the last PR. 
Forcing >1 parallelism would require changing CleanSparkSession to allow a new conf to override the default conf - not sure if that's worth it

python/src/spark_rapids_ml/umap.py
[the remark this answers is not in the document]
eordentlich: We pretty much require that to be enabled to get good data transfer from jvm to python workers.

python/src/spark_rapids_ml/umap.py
rishic3: Done, thx

eordentlich (approved):
👍

PR #833, opened 2025-02-04
for dataproc and databricks use ipython startup file to enable no-import [skip ci]

eordentlich:
to import spark_rapids_ml.install
note that in both cases, default NBs are run as root.

also use more parallelism for notebooks and example instances, leveraging stage level scheduling

### DIFF
--- a/notebooks/dataproc/README.md
+++ b/notebooks/dataproc/README.md
@@ -52,7 +52,7 @@ If you already have a Dataproc account, you can run the example notebooks on a D
   --metadata rapids-version=${RAPIDS_VERSION} \
   --metadata spark-rapids-ml-no-import-enabled=0 \
   --properties spark:spark.executor.resource.gpu.amount=1,\
-  spark:spark.task.resource.gpu.amount=1,\
+  spark:spark.task.resource.gpu.amount=0.0625,\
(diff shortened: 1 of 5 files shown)


eordentlich:
build

notebooks/dataproc/README.md
```diff
@@ -52,7 +52,7 @@ If you already have a Dataproc account, you can run the example notebooks on a D
   --metadata rapids-version=${RAPIDS_VERSION} \
   --metadata spark-rapids-ml-no-import-enabled=0 \
   --properties spark:spark.executor.resource.gpu.amount=1,\
-  spark:spark.task.resource.gpu.amount=1,\
+  spark:spark.task.resource.gpu.amount=0.0625,\
```
rishic3, line 55: Curious - why the discrepancy in parallelism between databricks/dataproc?

eordentlich: These are specific to the instances in the examples with dataproc ones having 16 cpu cores and emr and databricks 8 cores.   The setting should be 1/no cpu cores (though anything less will also be ok since the executor cores and task cpus settings will kick).

lijinf2 (approved):

PR #836, opened 2025-02-10
Add test case for LogisticRegressionModel.copy()

lijinf2:


lijinf2:
build

lijinf2:
build

python/tests/test_logistic_regression.py
```diff
@@ -368,15 +369,31 @@ def test_params(tmp_path: str, caplog: LogCaptureFixture) -> None:
...
+        _test_est_copy(LogisticRegression, input_spark_params, cuml_params_update)
+
+
+def test_lr_model_copy() -> None:
+
+    from .test_common_estimator import _test_model_copy
+    from .utils import get_toy_model
+
+    model_params: List[Dict[str, Any]] = [
+        {"featuresCol": "fea_dummy"},
```
eordentlich, line 386: Are these the only params that make sense to change for a model copy?

lijinf2: Yes, I think so. Other params (e.g. fit_intercept, regParam) are for fit only

PR #848, opened 2025-02-21
fix hang in pyspark-rapids shell

eordentlich:
broke while copy pasting.

Not clear how to test this script in ci.

cat script.py | pyspark doesn't work as expected and as a result neither does cat script.py | pyspark-rapids.


eordentlich:
build

lijinf2 (approved):

PR #842, opened 2025-02-18
Update pooling confs to use spark.rapids.memory.gpu.pool

rishic3:
Per [spark-rapids docs](https://nvidia.github.io/spark-rapids/docs/additional-functionality/advanced_configs.html), looks like `spark.rapids.memory.gpu.pooling.enabled` is deprecated in favor of `spark.rapids.memory.gpu.pool`.

PR #855, opened 2025-02-25
Revert "Update pooling confs to use spark.rapids.memory.gpu.pool"

rishic3:
Reverts https://github.com/NVIDIA/spark-rapids-ml/pull/842. rapids.daemon python workers do not respect the new pooling configs when `spark.rapids.sql.python.gpu.enabled=true` (see [this issue](https://github.com/NVIDIA/spark-rapids/issues/12228)); need to use old config to disable pooling.


eordentlich (approved):
👍

PR #869, opened 2025-03-13
Require spark.sql.execution.arrow.pyspark.enabled=true [skip ci]

rishic3:
Make this conf suggestion explicit, since UMAP relies on it to write attributes. See [this thread](https://github.com/NVIDIA/spark-rapids-ml/pull/823#discussion_r1914157957).


eordentlich (approved):
For future, let's check if we can enable this at runtime in our library have it take effect.

rishic3:
build

Issue #873, opened 2025-03-21
[bug] test_umap failed multiple cuda errors in nightly test intermittently


pxLi:
first seen in spark-rapids-ml_nightly, run: 666 (latest branch-25.04)

```
[2025-03-21T03:58:43.299Z] FAILED tests/test_umap.py::test_umap_sparse_vector[hamming-12-64-3000] - RuntimeError: CUDA error at: /root/miniconda3/include/rmm/device_uvector.hpp:325: cudaErrorIllegalAddress an illegal memory access was encountered
[2025-03-21T03:58:43.299Z] FAILED tests/test_umap.py::test_umap_sparse_vector[correlation-12-64-3000] - MemoryError: std::bad_alloc: CUDA error at: /root/miniconda3/include/rmm/mr/device/cuda_memory_resource.hpp
[2025-03-21T03:58:43.299Z] FAILED tests/test_umap.py::test_umap_precomputed_knn[tuple] - cupy_backends.cuda.libs.curand.CURANDError: CURAND_STATUS_INITIALIZATION_FAILED
```

from the conda deps comparing to previous successful cases, there is no significant changes of cuda-related pkgs.
```
...
[2025-03-20T17:01:20.791Z]     cuda-bindings-11.8.6       |  py310h629b23f_0         4.5 MB  conda-forge
[2025-03-20T17:01:20.791Z]     cuda-profiler-api-11.8.86  |                0          18 KB  nvidia
[2025-03-20T17:01:20.791Z]     cuda-python-11.8.6         |     pyha6e82b0_1          15 KB  conda-forge
[2025-03-20T17:01:20.791Z]     cuda-version-11.8          |       h70ddcb2_3          21 KB  conda-forge
[2025-03-20T17:01:20.791Z]     cudatoolkit-11.8.0         |      h4ba93d1_13       682.5 MB  conda-forge
[2025-03-20T17:01:20.791Z]     cudf-25.02.02              |cuda11_py310_250303_g8139f3c84f_0         1.3 MB  rapidsai
[2025-03-20T17:01:20.791Z]     cuml-25.02.01              |cuda11_py310_250303_g23773cd56_0         4.0 MB  rapidsai
[2025-03-20T17:01:20.791Z]     cupy-13.4.0                |  py310h386154f_0         348 KB  conda-forge
[2025-03-20T17:01:20.791Z]     cupy-core-13.4.0           |  py310h5da974a_0        43.4 MB  conda-forge
[...]

PR #877, opened 2025-03-26
Use randomized sparse data for UMAP testing

rishic3:
Aims to address #873 to avoid our previous pathological dataset, where all rows were largely duplicated. Added testing for both real-valued and binary sparse data for various metrics.

### DIFF
--- a/python/tests/test_umap.py
+++ b/python/tests/test_umap.py
@@ -41,29 +41,54 @@ from .utils import (
 )
 
 
-def _load_sparse_binary_data(
-    n_rows: int, n_cols: int, nnz: int
+def _load_sparse_data(
+    n_rows: int, n_cols: int, nnz: int, binary: bool = False
 ) -> Tuple[List[Tuple[SparseVector]], csr_matrix]:
-    # TODO: Replace this function by adding to SparseDataGen
-    # Generate binary sparse data compatible with Jaccard, with nnz non-zero values per row.
+    """
+    Generate random binary or real-valued sparse data with approximately nnz non-zeros per row.
+    """
+    density = nnz / n_cols
+    if binary:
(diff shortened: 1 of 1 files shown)


rishic3:
build

python/tests/test_umap.py
```diff
@@ -41,31 +41,61 @@
...
+    n_rows: int, n_cols: int, nnz: int, binary: bool = False
 ) -> Tuple[List[Tuple[SparseVector]], csr_matrix]:
-    # TODO: Replace this function by adding to SparseDataGen
-    # Generate binary sparse data compatible with Jaccard, with nnz non-zero values per row.
+    """
+    Generate random binary or real-valued sparse data with approximately nnz non-zeros per row.
+    """
+    density = nnz / n_cols
+    if binary:
+        data_rvs = lambda n: np.ones(n)
```
lijinf2, line 52: Looks good. 
Maybe keep the test case for testing the special case of all 0-distance neighbors, to track whether future release has taken care of the special case.

rishic3: We could, although that really seems like it should be a unit test on the cuml side given it's spark-independent. (Also the fix on the cuml side would probably just be erroring out). Wdyt?

lijinf2: One workaround is to add error message checking and expected error for "At least one row does not have any neighbor with non-zero distance" from 25.06 and forward.

lijinf2: error message copied from cuml PR https://github.com/rapidsai/cuml/pull/6491. 
Spark rapids ml just checked if the error message got printed out properly.

rishic3: Sure, will add this for 25.06

lijinf2 (approved):

Issue #879, opened 2025-04-01
[BUG] test_umap_sparse_vector[hellinger-12-64-3000] - AssertionError: assert 36000 == 0


pxLi:
first seen in park-rapids-ml_nightly run: 675

```
[2025-04-01T03:56:05.617Z] FAILED tests/test_umap.py::test_umap_sparse_vector[hellinger-12-64-3000] - AssertionError: assert 36000 == 0
[2025-04-01T03:56:05.617Z]  +  where 36000 = <Compressed Sparse Row sparse matrix of dtype 'float32'\n	with 36000 stored elements and shape (3000, 64)> != <Compressed Sparse Row sparse matrix of dtype 'float32'\n	with 36000 stored elements and shape (3000, 64)>.nnz
```

```

[2025-04-01T03:56:05.609Z] =================================== FAILURES ===================================

[2025-04-01T03:56:05.609Z] ________________ test_umap_sparse_vector[hellinger-12-64-3000] _________________

[2025-04-01T03:56:05.609Z] [gw2] linux -- Python 3.10.16 /root/miniconda3/bin/python3.10

[2025-04-01T03:56:05.609Z] 

[2025-04-01T03:56:05.609Z] n_rows = 3000, n_cols = 64, nnz = 12, metric = 'hellinger', gpu_number = 1

[2025-04-01T03:56:05.609Z] tmp_path = '/tmp/spark_rapids_ml_tests_yyjiaemj'

[2025-04-01T03:56:05.609Z] 

[2025-04-01T03:56:05.609Z]     @pytest.mark.parametrize("n_rows", [3000])

[2025-04-01T03:56:05.609Z]     @pytest.mark.parametrize("n_cols", [64])

[2025-04-01T03:56:05.609Z]     @pytest.mark.parametrize("nnz", [12])

[2025-04-01T03:56:05.609Z]     @pytest.mark.parametrize(

[2025-04-01T03:56:05.609Z]         "metric",

[2025-04-01T03:56:05.609Z]         [

[2025-04-01T03:56:05.609Z]             pytest.param("jaccard", id="jaccard"),

[2025-04-01T03:56:05.609Z]             pytest.param("euclidean", id="euclidean"),

[2025-04-01T03:56:05.609Z]             pytest.param("correlation", marks=pytest.mark.slow, id="correlation"),

[2025-04-01T03:56:05.609Z]             pytest.param("cosine", marks=pytest.mark.slow, id="cosine"),

[...]

PR #880, opened 2025-04-01
Fix UMAP test for sparse data w/Hellinger distance

rishic3:
Addresses #879. Fix mismatch in normalization for data provided to UMAP under hellinger metric. 
Normalization will change the order of data/indices in the CSR arrays, so check the dense versions match in such a case.

### DIFF
--- a/python/tests/test_umap.py
+++ b/python/tests/test_umap.py
@@ -42,16 +42,19 @@ from .utils import (
 
 
 def _load_sparse_data(
-    n_rows: int, n_cols: int, nnz: int, binary: bool = False
+    n_rows: int, n_cols: int, nnz: int, binary: bool = False, normalize: bool = False
 ) -> Tuple[List[Tuple[SparseVector]], csr_matrix]:
     """
     Generate random binary or real-valued sparse data with approximately nnz non-zeros per row.
+    If normalize is True, the data is normalized to have unit row sum (e.g., for hellinger distance).
     """
     density = nnz / n_cols
+
     if binary:
         data_rvs = lambda n: np.ones(n)
     else:
         data_rvs = lambda n: np.random.uniform(0.1, 1.0, n)
+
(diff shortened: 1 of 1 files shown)


python/tests/test_umap.py
```diff
@@ -42,16 +42,19 @@
 
 
 def _load_sparse_data(
-    n_rows: int, n_cols: int, nnz: int, binary: bool = False
+    n_rows: int, n_cols: int, nnz: int, binary: bool = False, normalize: bool = False
 ) -> Tuple[List[Tuple[SparseVector]], csr_matrix]:
     """
     Generate random binary or real-valued sparse data with approximately nnz non-zeros per row.
+    If normalize is True, the data is normalized to have unit row sum (e.g., for hellinger distance).
```
lijinf2, line 49: Seems the function does not guarantee nnz non-zeros per row if normalize is enabled. Maybe update the docstring to reflect that.

rishic3: Not sure i follow — normalization will preserve the nnz per row, only the values are scaled

lijinf2: oh, I find I was looking at the data gen process and thought the problem was there. 
So the PR is mainly fixing indices order mismatch of two csr matries (i.e. trained model raw data and input raw data)? Any idea on what changes the indice order of input raw data?

rishic3: Sorry should have clarified. Fixes are 1) passing the normalized data to create the spark dataframe and 2) accounting for order mismatch by comparing dense arrays. 
Can't point to the exact code but generally scipy's CSR operations [don't bother to preserve index orders](https://docs.scipy.org/doc/scipy-0.19.1/reference/sparse.html#further-details) as long as the content is correct.

lijinf2: Good observation!

lijinf2 (approved):

PR #883, opened 2025-04-08
Add unit test for connect ml plugin

wbo4958:
The added unit tests are not testing the end-to-end case for the 'connect'. Instead, the tests simply ensure that the functionality of the spark rapids ml JVM estimator/model could work even without  'connect' , behaving like a normal ML operator.


eordentlich (approved):
Nice.   Should we add to ci?

wbo4958:
> Nice. Should we add to ci?

Sounds good, @pxLi, could you help on it?

wbo4958:
build

pxLi:
> > Nice. Should we add to ci?
> 
> Sounds good, @pxLi, could you help on it?

this should be managed by developer directly as we have an agreement script at 
https://github.com/NVIDIA/spark-rapids-ml/blob/branch-25.06/ci/test.sh

please create a ticket to clarify the detailed requirement of the CI.
what SW Env should be involved (changes to the dockerfile), what commands should be covered.

PR #884, opened 2025-04-09
poc for testing init setter getter and enabling spark rapids ml to run with unsupported spark params

lijinf2:
Currently, if a spark_param (e.g. threshold of LogisticRegression) is mapped to None, spark rapids ml throws out an error and stops. If mapped to empty string "", spark rapids ml prints out a warning and continues. 

Do we let spark rapids ml run in both cases?


lijinf2:
build

eordentlich:
Let's keep current approach of failing hard in cases that map to None, but consistently for setters and constructor init.

PR #891, opened 2025-04-15
Add support read/write for connect plugin

wbo4958:


wbo4958:
build

wbo4958:
Hey @eordentlich, Could you help review it. Thx

jvm/src/main/scala/org/apache/spark/ml/rapids/RapidsLogisticRegressionModel.scala
```diff
@@ -50,4 +56,44 @@ class RapidsLogisticRegressionModel(override val uid: String,
...
+    new RapidsLogisticRegressionModel.RapidsLogisticRegressionModelWriter(this)
+}
+
+object RapidsLogisticRegressionModel extends MLReadable[RapidsLogisticRegressionModel] {
+
+  override def read: MLReader[RapidsLogisticRegressionModel] = new RapidsLogisticRegressionModelReader
+
+  override def load(path: String): RapidsLogisticRegressionModel = super.load(path)
+
+  private class RapidsLogisticRegressionModelWriter(instance: RapidsLogisticRegressionModel)
```
eordentlich, line 69: Looks like saved model won't be compatible with original spark-rapids-ml format.  Not sure we need that.   

Also, looks like it **is compatible** with spark mllib if plugin is disabled, it can be read in as spark mllib model.  Am I seeing that correctly?

jvm/src/main/scala/org/apache/spark/ml/rapids/RapidsLogisticRegressionModel.scala
```diff
@@ -50,4 +56,44 @@ class RapidsLogisticRegressionModel(override val uid: String,
...
+
+    override protected def saveImpl(path: String): Unit = {
+      val writer = instance.cpuModel.write
+      if (shouldOverwrite) {
+        writer.overwrite()
+      }
+      optionMap.foreach { case (k, v) => writer.option(k, v) }
+      writer.save(path)
+
+      val attributesPath = new Path(path, "attributes").toString
```
eordentlich, line 80: We should sync up with cpu fallback change PR.   Currently cpu fallback just returns a cpu model, no additional attributes.    Transform PR would be impacted as well.   Not sure that is what we want though.

eordentlich (approved):
Should go into 25.04 along with previous tranform PR backport.

jvm/src/main/scala/org/apache/spark/ml/rapids/RapidsLogisticRegressionModel.scala
wbo4958: Yeah, You're right. Let me fix it in a followup.

Issue #892, opened 2025-04-16
[BUG] test_make_sparse_regression[100-None-True-1000-200-density1-True-5-bias1-0-False-float64] - assert 4 == 5


pxLi:
spark-rapids-ml_nightly, run: 689

```
[2025-04-16T04:23:27.123Z] =================================== FAILURES ===================================

[2025-04-16T04:23:27.123Z] _ test_make_sparse_regression[100-None-True-1000-200-density1-True-5-bias1-0-False-float64] _

[2025-04-16T04:23:27.123Z] [gw2] linux -- Python 3.10.17 /root/miniconda3/bin/python3.10

[2025-04-16T04:23:27.124Z] 

[2025-04-16T04:23:27.124Z] dtype = 'float64', use_gpu = 'False', redundant_cols = '0'

[2025-04-16T04:23:27.124Z] logistic_regression = 'True', n_classes = '5'

[2025-04-16T04:23:27.124Z] bias = ['0.5', '1.5', '2.5', '3.5', '4.5'], density = ['0.05', '0.1', '0.2']

[2025-04-16T04:23:27.124Z] rows = '1000', cols = '200', density_curve = 'None', shuffle = 'True'

[2025-04-16T04:23:27.124Z] n_chunks = '100'

[2025-04-16T04:23:27.124Z] 

[2025-04-16T04:23:27.124Z]     @pytest.mark.parametrize("dtype", ["float64"])

[2025-04-16T04:23:27.124Z]     @pytest.mark.parametrize("use_gpu", ["True", "False"])

[2025-04-16T04:23:27.124Z]     @pytest.mark.parametrize("redundant_cols", ["0", "2"])

[2025-04-16T04:23:27.124Z]     @pytest.mark.parametrize(

[2025-04-16T04:23:27.124Z]         "logistic_regression, n_classes, bias",

[2025-04-16T04:23:27.124Z]         [

[2025-04-16T04:23:27.124Z]             ("True", "2", "1.0"),

[2025-04-16T04:23:27.124Z]             ("True", "5", ["0.5", "1.5", "2.5", "3.5", "4.5"]),

[2025-04-16T04:23:27.124Z]             ("True", "15", "1.5"),

[2025-04-16T04:23:27.124Z]             ("False", "0", "1.0"),

[2025-04-16T04:23:27.124Z]         ],

[2025-04-16T04:23:27.124Z]     )

[...]

PR #894, opened 2025-04-17
Make test_make_sparse_regression more stable and explain the non-determinism of SparseRegressionDataGen

lijinf2:
Issue: https://github.com/NVIDIA/spark-rapids-ml/issues/892
Relevant PR: https://github.com/NVIDIA/spark-rapids-ml/pull/742

### DIFF
--- a/python/benchmark/gen_data_distributed.py
+++ b/python/benchmark/gen_data_distributed.py
@@ -580,6 +580,11 @@ class RegressionDataGen(DataGenBaseMeta):
 class SparseRegressionDataGen(DataGenBaseMeta):
     """Generate sparse regression dataset using a distributed version of sklearn.datasets.regression,
     including features and labels.
+
+    SparseRegressionDataGen is a non-deterministic data generator that may produce different DataFrames across multiple runs.
+    While setting a random state ensures that the generated DataFrame will vary between runs, it does not guarantee reproducibility of the same output for the same random state.
+    This is because each partition utilizes a combination of the random state and the partition index as its local seed.
+    However, partition indices are not guaranteed to be continuous or deterministic, leading to variability even when the random state is provided.
     """
 
     def __init__(self, argv: List[Any]) -> None:
--- a/python/benchmark/test_gen_data.py
+++ b/python/benchmark/test_gen_data.py
@@ -264,6 +264,10 @@ def test_make_sparse_regression(
     shuffle: str,
     n_chunks: str,
 ) -> None:
+    """
+    SparseRegressionDataGen is a non-deterministic data generator and may produce different DataFrames even when initialized with the same random state.
+    For more details, please refer to the class docstring.
+    """
     input_args = [
         "--num_rows",
         rows,
--- a/python/benchmark/test_gen_data.py
+++ b/python/benchmark/test_gen_data.py
@@ -238,7 +238,7 @@ def test_make_regression(
 )
 @pytest.mark.parametrize(
     "rows, cols",
-    [("1000", "200"), pytest.param("10000", "1000", marks=pytest.mark.slow)],
+    [("10000", "200"), pytest.param("10000", "1000", marks=pytest.mark.slow)],
 )
 @pytest.mark.parametrize(
     "density_curve, shuffle",


lijinf2:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -580,6 +580,11 @@ def make_regression_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
 class SparseRegressionDataGen(DataGenBaseMeta):
     """Generate sparse regression dataset using a distributed version of sklearn.datasets.regression,
     including features and labels.
+
+    SparseRegressionDataGen is a non-deterministic data generator that may produce different DataFrames across multiple runs.
+    While setting a random state ensures that the generated DataFrame will vary between runs, it does not guarantee reproducibility of the same output for the same random state.
+    This is because each partition utilizes a combination of the random state and the partition index as its local seed.
+    However, partition indices are not guaranteed to be continuous or deterministic, leading to variability even when the random state is provided.
```
eordentlich, line 587: We have seen they can be non-contiguous.   Why not deterministic?

lijinf2: Printed out generated y non-deterministic. 
Need further investigation.

lijinf2:
build

python/benchmark/gen_data_distributed.py
eordentlich: Good catch.   So is it still non-deterministic?

python/benchmark/gen_data_distributed.py
```diff
@@ -580,6 +592,12 @@ def make_regression_udf(iter: Iterable[pd.DataFrame]) -> Iterable[pd.DataFrame]:
 class SparseRegressionDataGen(DataGenBaseMeta):
     """Generate sparse regression dataset using a distributed version of sklearn.datasets.regression,
     including features and labels.
+
+    SparseRegressionDataGen is a deterministic data generator when running in Spark local mode. It produces
+    identical DataFrames when initialized with the same random state. However, in Spark standalone mode, it may occasionally
+    yield different DataFrames even with the same random state. This is because each partition derives its seed from
+    a combination of the provided random state and the partition index. Since partition indices are not guaranteed to be continuous
+    or consistent across runs in standalone mode, this can lead to non-deterministic output despite a fixed random state.
```
eordentlich, line 600: Should be consistent as long as number of partitions is the same but that could change depending on number of executor-cores for example.    Maybe need to give option to set the num partitions argument to spark.range , the base dataframe/rdd.

lijinf2:
build

python/benchmark/gen_data_distributed.py
lijinf2: Revised the comment.

python/benchmark/gen_data_distributed.py
lijinf2: Added a warning message when output_num_files not specified.

python/benchmark/gen_data_distributed.py
```diff
@@ -785,7 +802,6 @@ def make_sparse_regression_udf(
                     random_state=generator,
                     format="csr",
                     dtype=dtype,
-                    data_rvs=np.random.randn,
```
eordentlich, line 788: Curious why this was deleted?

python/benchmark/gen_data_distributed.py
```diff
@@ -816,10 +831,7 @@ def make_sparse_regression_udf(
                     sparse_matrix = sparse_matrix[:, informative_shuffle_indices]
                 informative = sparse_matrix[:, :n_informative]
 
-                if use_cupy:
-                    redundant_mul = np.random.rand(n_informative, redundant_cols)
```
eordentlich, line 820: Was this a bug to use np vs cp here?

lijinf2: Not a bug. redundant_mul must be a numpy array because it is used to construct a CPU-based SciPy sparse matrix, regardless of whether use_cupy is true or false. cupy is only utilized for generating the label column.

The cupyx sparse matrix is not used here, which may be fine for now.

python/benchmark/gen_data_distributed.py
lijinf2: For the function to reuse generator which was initialized with random_state. 
Keeping it here will instruct the function to use np.random.randn, which creates non-determinism and it was not initialized with random_state.

eordentlich: Does the distribution change from normal to uniform 0,1?  To keep distribution could an initialized generator function be passed in here?

lijinf2: Good catch. 

Just added back the data_rvs per np.random had been revised to be initialized with the partition seed.

lijinf2:
build

python/benchmark/gen_data_distributed.py
```diff
@@ -807,7 +824,6 @@ def make_sparse_regression_udf(
                         random_state=generator,
                         format="csr",
                         dtype=dtype,
-                        data_rvs=np.random.randn,
```
eordentlich, line 810: Here too looks like.

lijinf2:
build

eordentlich (approved):
👍

PR #897, opened 2025-04-22
Accelerating the pipeline by eliminating vector assembly for columnar input data in Spark Rapids ML

lijinf2:


lijinf2:
build

lijinf2:
build

python/tests/test_pipeline.py
```diff
@@ -0,0 +1,404 @@
...
+        (Pipeline, RandomForestRegressor, RandomForestRegressionModel, {"maxDepth": 3}),
+        (
+            SparkPipeline,
+            SparkRandomForestRegressor,
+            SparkRandomForestRegressionModel,
+            {"maxDepth": 3},
+        ),
+    ],
+)
+def test_example_random_forest(
```
rishic3, line 396: Presumably the pipeline works for our non-spark-mllib algos (UMAP, DBSCAN, etc.) as well right? Could we add a test for one or more?

lijinf2:
build

python/tests/test_pipeline.py
lijinf2: Added UMAP, DBSCAN to tests.
Also revised umap to support n_components=1 when cuml obj returns pd.Series instead of pd.DataFrame. 
Did not add KNN/ANN to tests as pipeline does not work with kneighbors API.

python/src/spark_rapids_ml/pipeline.py
```diff
@@ -0,0 +1,108 @@
...
+
+        for i in range(len(stages)):
+
+            if not self._isGPUEstimator(stages[i]):
+                continue
+
+            if (
+                i > 0
+                and isinstance(stages[i - 1], VectorAssembler)
+                and self._allScalar(dataset, stages[i - 1].getInputCols())  # type: ignore
```
eordentlich, line 67: Does this actually work in the general case of arbitrary Pipelines which can implement almost arbitrary DAGs since input columns of a stage can be output of any previous stage or the pipeline input.

python/src/spark_rapids_ml/pipeline.py
```diff
@@ -0,0 +1,108 @@
...
+            ):
+
+                self._setInputCols(stages[i], stages[i - 1].getInputCols())  # type: ignore
+                stages[i - 1] = NoOpTransformer()
```
eordentlich, line 71: Does this mean NoOpTransformer will be part of the PipelineModel returned?  

If so, that gets a bit tricky since we won't match Spark pipeline.transform output which  I believe has all stage output columns, in this case the VectorAssembler output column as well.   Pipeline Transform seems challenging to support with this trick for now.

python/src/spark_rapids_ml/pipeline.py
```diff
@@ -0,0 +1,108 @@
...
+        from spark_rapids_ml.utils import setInputOrFeaturesCol
+
+        setInputOrFeaturesCol(gpu_est, input_cols)
+
+    @staticmethod
+    def _allScalar(dataset: DataFrame, inputCols: List[str]) -> bool:
+        from pyspark.sql.types import NumericType
+
+        for col_name in inputCols:
+            if not isinstance(dataset.schema[col_name].dataType, NumericType):
```
eordentlich, line 94: In general case, col_name might not be a column of dataset.   Would that raise an exception?

python/tests/test_pipeline.py
```diff
@@ -0,0 +1,436 @@
...
+
+    algo_params = {
+        "featuresCol": "features",
+        "labelCol": label_col,
+        "maxIter": 10,
+        "regParam": 0.001,
+    }
+    est = Est(**algo_params)
+
+    pipeline = PipelineEst(stages=[assembler, est])
```
eordentlich, line 110: Any tests with more than two stages?   Pipelines can be very complicated DAGs in the general case through column references to preceding stage outputs.   See comments above.

eordentlich (commented):
This is implemented as if it works for all Pipelines but only tested for simple 2 stage pipelines.  Not sure it actually works in the general case of arbitrary pipelines.

python/src/spark_rapids_ml/pipeline.py
lijinf2: Yeah. The PR only works for (vec_assembler, gpu_estimator) pairs in the list of stages, and it does not work for other patterns.

python/src/spark_rapids_ml/pipeline.py
lijinf2: ok. 

It should be possible to put vec_assembler back to the stage list after Estimator fitting. Will revise.

python/src/spark_rapids_ml/pipeline.py
lijinf2: Regarding (vec_assembler, gpu_estimator) pairs I meant adjacent pairs, so pipeline would look like (..., vec_assembler, gpu_estimator, ...)

eordentlich: Example:  vec_assembler,gpu_logistic,cpu_gbdt  where cpu_gpdt also uses output of vec assembler.  so this pipeline (after fit and transform) adds 3 new columns to input df: the output of vec_assembler, logistic prediction and cpu gbdt prediction.   I believe this is a valid pipeline even though it has two fit/transform stages relying on the same input.   There could be another stage that ensembles these two outputs.   The current code would break on this since cpu_gbdt would be getting columnar input and not vector.   

The point is that most generally pipelines can be arbitrary dags with each stage fit/transform adding new columns and arbitrary references to columns in previous stages as inputs.

python/src/spark_rapids_ml/pipeline.py
lijinf2: revised to call spark pipeline in this case.

python/tests/test_pipeline.py
lijinf2: Make sense. Now PR revised to handle two stages only. 
Need to investigage the case if more than two stages.

lijinf2:
build

lijinf2:
build

eordentlich (approved):
👍

PR #903, opened 2025-04-23
Add a messaging suggesting replacing ivf_pq with nn_descent for cagra when error happens

lijinf2:
nn_descent seems to be more stable at the current stage.


lijinf2:
build

eordentlich (approved):
👍

PR #925, opened 2025-05-19
Support kmeans for connect plugin

wbo4958:


wbo4958:
build

eordentlich (approved):
👍 
General question about transferring both the model to java via cpu() and the attributes json.   Are both needed?   this would apply to all the algos?   
Also, curious if cpu() for the case of large/many KMeans cluster centers decently fast?

wbo4958:
> General question about transferring both the model to java via cpu() and the attributes json. Are both needed? this would apply to all the algos?

Actually, I think only model attributes of spark-rapids-ml is enough. The corresponding CPU model is constructed by the model attributes. Transferring both (java model and attributes) was for fast POC, or else, we need to construct java model in the JVM layer.

> Also, curious if cpu() for the case of large/many KMeans cluster centers decently fast?
It's going to convert the cluster centers to java vector which could be slow if it's large cluster centers.

PR #929, opened 2025-05-23
Use int64 instead of int32 index dtype for sparse logistic regression on large dataset to avoid array index overflow

lijinf2:
int32 theoretically supports up to 2.1 billion nnz. But some cuda kernel (e.g. map_kernel) calculates array index which can be overflow if int32 is used.

Let int32 support up to 1 billion nnz and switch to int64 if nnz is larger than 1 billion.


lijinf2:
build

eordentlich (commented):
👍

lijinf2:
build

eordentlich (approved):
👍

PR #930, opened 2025-05-26
Support array input for connect plugin

wbo4958:
This PR tries to override `transformSchema` function in estimator which is going to check if the feature is Vector input, if not, this function is going to throw exception. Given that, spark-rapids-ml supports vector/array/multicolumns, we need to eliminate this checking to make it support array input.


wbo4958:
build

wbo4958:
> 👍
> 
> Is fit() impacted by this?

The spark ml fit couldn't handle Array input, it is transformSchema where to throw exception. So we override it and delegate the checking to the spark-rapids-ml.

PR #933, opened 2025-05-28
Construct cpu model in JVM side for connect plugin

wbo4958:
It's super slow when constructing the corresponding CPU Model in python especially for RandomForest cases which have many trees. This PR moves the cpu model construction from python side to jvm side.


eordentlich (approved):
👍  This should avoid the observed slow down.   And nice update to official spark 4.0.0 mvn dep.

wbo4958:
build

PR #936, opened 2025-05-29
Enable pipeline to use vectorassembler when handleinvalid is not error

lijinf2:


lijinf2:
build

python/src/spark_rapids_ml/tree.py
```diff
@@ -547,10 +547,12 @@ def featureImportances(self) -> Vector:
         """Estimate the importance of each feature."""
         return self.cpu().featureImportances
 
-    @property
-    def getNumTrees(self) -> int:
-        """Number of trees in ensemble."""
-        return self.getOrDefault("numTrees")
+    if not TYPE_CHECKING:
```
eordentlich, line 550: If everything is type ignore, is this still needed?

lijinf2: The root cause seems to be latest mypy does not differentiate function definition and property definition (Spark RandomForestModel object has a getNumTree property). 

TYPE_CHECKING is needed to avoid mypy checking this property definition. Otherwise, mypy throws out a "definition incompatible" error. 

Sorry I did not follow "everything is type ignore". May I have more details?

lijinf2:
build

python/src/spark_rapids_ml/tree.py
eordentlich: But good to add a comment/TBD to indicate this in the code.

eordentlich (approved):
let's revisit the mypy exception later.  ok for now to unblock ci.  rest looks good.

PR #951, opened 2025-06-17
Print out a warning message for very wide sparse dataset when converting it to dense for involking cuml

lijinf2:


lijinf2:
build

python/src/spark_rapids_ml/core.py
```diff
@@ -508,6 +510,12 @@ def _pre_process_data(self, dataset: DataFrame) -> Tuple[
                     assert use_sparse is False or (
                         use_sparse is None and first_vectorudt_type is DenseVector
                     )
+
+                    if dimension > 10_000:
```
eordentlich, line 514: Why 10000?

python/src/spark_rapids_ml/core.py
```diff
@@ -508,6 +510,12 @@ def _pre_process_data(self, dataset: DataFrame) -> Tuple[
...
+                        get_logger(self.__class__).warning(
+                            "Spark RAPIDS ML converts sparse vectors to dense format for algorithms that do not natively support sparse input in cuML. This conversion can cause the data size per GPU to exceed Arrow’s maximum supported size (MAX_INT). If Arrow raises an exception, consider reducing the number of rows or the vector dimensionality per GPU."
```
eordentlich, line 516: This warning needs some clarification.

Is it necessarily true that sparse vectors were being used?   

Also what is 'maximum supported size'?   Bytes?

And 'vector dimensionality per GPU' can't be controlled.

python/src/spark_rapids_ml/core.py
lijinf2: Revised.

python/src/spark_rapids_ml/core.py
lijinf2: Revised, please take another look.

lijinf2: Make the message as a warning message is the actual runtime error detection requires a count(). 

it should also work for sparse, multi_cols, and transform. Did not test thoroughly. Think just a warning message so it will not affect any runtime.

lijinf2: Seems same problem exists in transform too. lazy execution delays mapInPandas execution to application code.

eordentlich: Thinking more about this, since it seems impossible to cover all cases in our code, should we just note this issue and possible fix on a troubleshooting page or extend the performance tuning page to performance tuning & troubleshooting:
https://nvidia.github.io/spark-rapids-ml/

lijinf2: Good idea. Added a note to FAQ.md regarding the issue.

eordentlich (approved):
👍

PR #953, opened 2025-06-25
allow setting verbose, float32_inputs, and num_workers using spark confs

eordentlich:
This will allow setting these when connecting to Spark Connect server w/ spark-rapids-ml plugin without app code change.


eordentlich:
build

Issue #957, opened 2025-06-30
25.08 drop cuda11 support


pxLi:
As rapids stops supporting cuda11, 
lets update the corresponding CI and script to build and test against cuda 12 (cuda12.9.1 runtime) only after we switch dev default to rapids 25.08 version

1. spark-rapids-ml repo (code, CI premerge, doc)
2. internal jenkins repo (CI nightly)

PR #959, opened 2025-07-01
drop cuda11 support

YanxuanLiu:
fix https://github.com/NVIDIA/spark-rapids-ml/issues/957

- Updated default cuda version to 12.9.1
- TODO: `notebooks/databricks/init-pip-cuda-11.8.sh`

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG CUDA_VERSION=12.9.1
+ARG CUDA_VERSION=12.0.1
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04
 
 # ubuntu22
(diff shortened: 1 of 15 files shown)


ci/Dockerfile
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG CUDA_VERSION=11.8.0
+ARG CUDA_VERSION=12.9.1
```
eordentlich, line 17: Should we test with lowest cuda 12 version supported by rapids?  That would be 12.0 .   Not yet seeing 12.9 [officially supported](https://docs.rapids.ai/install/#selector) but I guess 25.08 will have that.

eordentlich:
Thank you @YanxuanLiu !  I will take another look but seems pretty comprehensive.   I will also test the databricks scripts before merging.

YanxuanLiu:
build

eordentlich (approved):
Looks good.   Will test databricks and other CSPs in follow up.

PR #973, opened 2025-08-13
fix gpu assignment in local-cluster mode and bash cli's

eordentlich:
Fixes gpu assignment in local-cluster mode (used in test of no-code change cli's) to match what is done local mode.

This gpu assignment error was silently failing in the cli runs in run_test.sh in CI runs because the cli's weren't propagating Spark errors to their exit codes.   This PR also fixes this issue in the clis by setting `check=True` in the Python `subprocess.run` invocations and adds a test to check the Spark errors are propagated correctly.


eordentlich:
build

rishic3 (approved):

PR #701, opened 2024-08-05
Support System Allocated Memory (SAM)

rongou:
Allowing running benchmarks using System Allocated Memory (SAM) on HMM and Grace Hopper systems. This depends on https://github.com/cupy/cupy/pull/8442.

PR #977, opened 2025-09-03
Support System Allocated Memory (SAM)

eordentlich:
Rebases https://github.com/NVIDIA/spark-rapids-ml/pull/701 to spark-rapids-ml 25.08 and cupy 13.6

Description from #701 :
Allowing running benchmarks using System Allocated Memory (SAM) on HMM and Grace Hopper systems. This depends on https://github.com/cupy/cupy/pull/8442.

Some additional changes:
- moved custom aligned numpy allocator config to core library and into a separate module (invoking in a function leads to segfault in cuml import likely due to a key object being deallocated but haven't root caused)
- set `CUPY_ENABLE_UMP` env var automatically in python worker run time if sam is enabled (i.e. no separate config needed)
- minor patches to benchmarking scripts (e.g. to fix verbose mode). 
- for algo's that concatenate input batches, memadvise numpy allocations to stay on host.  Otherwise, these migrate to device during concatenation and can lead to OOM wrt headroom calculation
- reserve device memory for sam rmm allocations during cuml fit, in addition to non-rmm allocations by libraries (this speeds up kmeans cuml fit by 13x over original making it 4x faster than uvm)

### DIFF
--- a/python/src/spark_rapids_ml/clustering.py
+++ b/python/src/spark_rapids_ml/clustering.py
@@ -509,7 +509,7 @@ class KMeansModel(KMeansClass, _CumlModelWithPredictionCol, _KMeansCumlParams):
         def _construct_kmeans() -> CumlT:
             from cuml.cluster.kmeans import KMeans as CumlKMeans
 
-            kmeans = CumlKMeans(output_type="cudf", **cuml_alg_params)
+            kmeans = CumlKMeans(output_type="cupy", **cuml_alg_params)
             from spark_rapids_ml.utils import cudf_to_cuml_array
 
             kmeans.n_features_in_ = n_cols


python/src/spark_rapids_ml/utils.py
```diff
@@ -156,20 +156,46 @@ def _get_gpu_id(task_context: TaskContext) -> int:
...
+            "Both CUDA managed memory and system allocated memory cannot be enabled at the same time."
+        )
+
+    if sam_enabled and sam_headroom is None:
+        if not type(rmm.mr.get_current_device_resource()) == type(
+            rmm.mr.SystemMemoryResource()
+        ):
+            mr = rmm.mr.SystemMemoryResource()
+            rmm.mr.set_current_device_resource(mr)
+    elif sam_enabled and sam_headroom is not None:
```
rishic3, line 179: Following @lijinf2's suggestion [here](https://github.com/NVIDIA/spark-rapids-ml/pull/701#discussion_r1704714084), i wonder if we can check first if SAM is supported, e.g. https://github.com/rapidsai/rmm/blob/7cef2f5f30e962e9f3b27a3a3f2753a40277c093/python/rmm/rmm/tests/test_rmm.py#L36

docs/site/configuration.md
```diff
@@ -6,14 +6,14 @@ nav_order: 6
...
 | spark.rapids.ml.float32_inputs | None | if set to a boolean value (true/false), controls whether input data should be converted to float32 precision before being passed to cuML algorithms. Setting this to true can reduce memory usage and potentially improve performance, but may affect numerical precision. This parameter can be set globally in Spark configuration and will be used if not explicitly set in the estimator constructor. |
 | spark.rapids.ml.num_workers | None | if set to an integer value greater than 0, specifies the number of workers to use for distributed training. This parameter can be set globally in Spark configuration and will be used if not explicitly set in the estimator constructor. |
 
-
-Since the algorithms rely heavily on Pandas UDFs, we also require `spark.sql.execution.arrow.pyspark.enabled=true` to ensure efficient data transfer between the JVM and Python processes.
```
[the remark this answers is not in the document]
eordentlich: No.  Good catch.

eordentlich:
build

python/src/spark_rapids_ml/classification.py
```diff
@@ -1087,6 +1088,24 @@ def _single_fit(init_parameters: Dict[str, Any]) -> Dict[str, Any]:
                 if is_sparse and pdesc.partition_max_nnz > nnz_limit_for_int32:  # type: ignore
                     logistic_regression._convert_index = np.int64
 
+                # if enabled, reduce same reserved memory to targeted amount
```
rishic3, line 1091: same -> sam (here and in other algos)

python/src/spark_rapids_ml/core.py
```diff
@@ -717,7 +760,18 @@ def _train_udf(pdf_iter: Iterator[pd.DataFrame]) -> pd.DataFrame:
...
+            # e.g. for kmeans cluster centers
+            _configure_memory_resource(
+                cuda_managed_mem_enabled,
+                cuda_system_mem_enabled,
+                (
+                    2 * cuda_system_mem_headroom
+                    if cuda_system_mem_headroom is not None
+                    else None
+                ),
+            )
```
rishic3, line 774: So if sam headroom is None is this a no-op? Since we are using the headroom to protect some RMM space for auxiliary objects seems we should almost require it to be non-zero?

rishic3, line 774: I guess 2x works for KMeans where auxiliary object size is roughly constant, but wondering if it won't be enough for algos like UMAP where auxiliary objects scale with data size.

rishic3:
Wonder how this interacts with gpuMemRatioForData. Does [device.mem_info](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-25.10/python/src/spark_rapids_ml/utils.py#L315) account for headroom, especially if headroom further gets padded after device memory is queried? I suppose if not it's fine as long as it's clear.

python/src/spark_rapids_ml/core.py
eordentlich: They will still run but perhaps not optimally.   Not clear how to optimally layout data size dependent structures, whether all or part should be in cpu vs the data itself.   Let's revisit on an algorithm by algorithm basis in future prs if there is interest.

rishic3: Right, so no RMM space is explicitly reserved for cuml fit objects? Not that it needs to change in this PR but just wondering if library headroom and cuML auxiliary headroom should be logically separate, e.g. we still might want the latter if former is off.

eordentlich: Might make sense, but opted to not add another config till the need is clear.   Added a TODO comment in the code.

eordentlich: Added a TODO on this.

eordentlich:
build

lijinf2:
> > Wonder how this interacts with gpuMemRatioForData. Does [device.mem_info](https://github.com/NVIDIA/spark-rapids-ml/blob/branch-25.10/python/src/spark_rapids_ml/utils.py#L315) account for headroom, especially if headroom further gets padded after device memory is queried? I suppose if not it's fine as long as it's clear.
> 
> With SAM, due to the check here not taking into account the headroom, the behavior would have been to potentially include host memory in the preallocated concatenation array but this would be fine with SAM and ATS, etc. But seemed like in this case, we should actually allow the concatenation array to span host and device memories, so made the change to have the specified gpuMemRatioForData multiply (available host + device memory).

Did we get a chance to test the gpuMemRatioForData (spanning host and device memories) on Grace-Blackwell machines (e.g. DGX Spark)? "free_mem, total_mem = device.mem_info" may already include the host memory. Would "free_mem += psutil.virtual_memory().available" lead to an overestimate of the available memory?

eordentlich:
Tested on grace-hopper where host and device memory are separate with distinct sizes.  on DGX-Spark like systems with a single unified memory, out-of-core is not really a valid concept, so no need for SAM mode and gpuMemRatioForData will just use whatever device info returns for memory.

eordentlich:
There is no reason to enable SAM on such a system.    Also DGX Spark is not even out yet so not urgent.

PR #979, opened 2025-09-15
Drop CUDA 11 support in pyproject [skip ci]

rishic3:
Since cuML dropped CUDA 11 support as of 25.08 seems like we should drop this from our pyproject for consistency.


rishic3:
build

YanxuanLiu (approved):

PR #982, opened 2025-09-30
updates for RAPIDS 25.10 compatibility

eordentlich:
- keeps up with on-going refactoring of cuml internal apis we invoke
- conversion of random forest models to spark mllib is now via treelite json format as cuml json format was removed
- ucx-py -> ucxx required some changes (currently knn and dbscan which rely on ucxx will yield incorrect results until https://github.com/rapidsai/raft/pull/2829 is merged and pulled in through cuml).


eordentlich:
build

